PR #9, opened 2024-08-10
Admin commands list update and a translation fix for Spanish

Jeii20:


Jeii20:
I made an accidental change at string BATTLE ROYALE, but it's fixed in the next file.

CritiKangel (approved):

CritiKangel:
You deleted the command "!guests" in string number 1354, after "!hidden!

Jeii20:
Yes, because the command doesn't exist anymore, I tested before I applied, and anyway they can't play without signing in! ^^

Jeii20:
> You deleted the command "!guests" in string number 1354, after "!hidden!

Yes, because the command doesn't exist anymore, I tested before I applied, and anyway they can't play without signing in! ^^

CritiKangel:
> > You deleted the command "!guests" in string number 1354, after "!hidden!
> 
> Yes, because the command doesn't exist anymore, I tested before I applied, and anyway they can't play without signing in! ^^

let's not delete anything from the original without the dev's permission

Jeii20:
> > > You deleted the command "!guests" in string number 1354, after "!hidden!
> > 
> > Yes, because the command doesn't exist anymore, I tested before I applied, and anyway they can't play without signing in! ^^
> 
> let's not delete anything from the original without the dev's permission

Noted, sorry for my mistake and thank you for letting me aware. ^^

PR #18, opened 2024-08-10
Update strings.xml

Briixss:
Errores ortográficos

ID: 23155118


Jeii20:
** spanish string

Majaaaa1:
@Briixss tag granted, thanks for your effort.

PR #33, opened 2024-08-12
Update ‏Turkish language improvements. xml

xrqqo:
Name: Ruqaya
ID: 22336157


zeynpcna2:
I have already made these corrections, you can check the current Tr file

PR #40, opened 2024-08-13
Translations for the new strings in italian + corrections

ghost:


Jeii20:
Belongs to `values-it`.

PR #44, opened 2024-08-14
Update strings.xml Arabic rami

3000RamiNebulous:
Translated 2X skin resolution line.
Made sure that there isn't any mistake.


3000RamiNebulous:
So basically I'm approved and should receive the things?

PR #73, opened 2024-08-16
Update strings.xml

Teiko33:
Hi here are some corrections and translations from the french language :

1.Mass boost translated to "Boost de Masse"

2.I replaced the word "pour" with "de". This means choose the color OF your level. The old translation "choose the color "FOR" your level is not appropriate

3.Invites translated to "invitations"

4.Complete campaign mission is a mission that we must do and not a mission that we have completed. The old translations were translated as a mission that we completed. So i corrected the translations by "Compléter la mission de campagne"

5.Activate translated to "activer"

6.Account must be 1 day old te send mail translated to "Le compte doit avoir 1 jour pour pouvoir envoyer un mail".

Teiko


Teiko33:
I already have it but thanksLe 16 août 2024 à 23:31, Greg ***@***.***> a écrit :﻿
Thanks, reply with your account ID if you want the translator tag and don't already have it.

PR #76, opened 2024-08-16
Update values-ja (Japanese) [3]

Jeii20:
Part 3: Translated more words and server names.


Jeii20:
I thought I missed a typo at string "Valentine's Day", but it's actually fine, that's why I quickly set as draft.

PR #80, opened 2024-08-17
Update values-vi (Vietnamese)

Jeii20:
"Translator" tag translated, commands list translated and added 4 other existing commands.


Jeii20:
Use the latest file, a typo regarding my translation has been corrected at the "Translator" string.

PR #83, opened 2024-08-17
Update strings.xml

W0O13:
I changed word مجموعه to مجموعة because it has different meaning I changed word ان to أن beacuse it writes like that in arabic I changed word غرفه to غرفة beacuse it writes like that in arabic i changed sentence تجعل خصمُك من الفرق سيؤدي ذلك الى حظرك من اللعبة to جعل خصمُك من الفريق سيؤدي ذلك إلى حظرك من اللعبة becuase the pre-written sentence has a completely inappropriate moral meaning,, as if it were a compound sentence


W0O13:
My name is: WOO
My id is : 11111113
and if it is done can i get the tag
Thanks.

PR #85, opened 2024-08-17
Indonesian translate

Vynits:


Vynits:
What should I do to get Tag translate

Pada Sen, 19 Agt 2024 9.52 AM, Greg ***@***.***> menulis:

fahmibql:
The Language Translation is weird , and it make it worse to see

PR #92, opened 2024-08-18
Update strings.xml Polish Improvments

esgamerx:
I fixed some problems in Polish and replaced English words with Polish.


esgamerx:
what about  the tag it looks like accepted translation .
my neb id : 19937636
@grog444

PR #105, opened 2024-08-19
Update vietnamse br language (example)

toitenxyzzz:


res/values-pt-rBR/strings.xml
```diff
@@ -59,7 +59,7 @@
     <string name="Game_Full">"Jogo Cheio"</string>
     <string name="US_East">"América Central"</string>
     <string name="US_West">"América do Norte"</string>
-    <string name="Europe">"Europa"</string>
+    <string name="Europe">"Europa."</string>
```
Jeii20, line 62: Looks like you completely based off on the mobile tutorial video. You placed an unnecessary period at the server name, and according to the title you provided, you should edit the `values-vi`, that belongs to Vietnamese language.

PR #107, opened 2024-08-19
Update strings.xml

Teiko33:
Some corrections of translations

The word "COURRIEL" is correct by using google translation however in France few people use it, it is a commonly used word, this why i prefer that we put "MAIL" instead it is a word often used. The word is already translated as MAIL so it's better to leave it like that

"resolution de la skin" is false, i don't know who translated it but it's not said at all

The word US was translated as America ( US east was translated as central america and west as north america i don't know why ), i corrected that.

### DIFF
--- a/res/values-fr/strings.xml
+++ b/res/values-fr/strings.xml
@@ -58,8 +58,8 @@
     <string name="SERVER">"SERVEUR"</string>
     <string name="Joining_game_in_">"Rejoindre la partie "</string>
     <string name="Game_Full">"Partie complète"</string>
-    <string name="US_East">"Amérique Centrale"</string>
-    <string name="US_West">"Amérique du Nord"</string>
+    <string name="US_East">"Etats-Unis de l'Est"</string>
+    <string name="US_West">"Etats-Unis de l'Ouest"</string>
     <string name="Europe">"Europe"</string>
     <string name="South_America">"Amérique du Sud"</string>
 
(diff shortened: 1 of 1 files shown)


res/values-fr/strings.xml
```diff
@@ -58,8 +58,8 @@
     <string name="SERVER">"SERVEUR"</string>
     <string name="Joining_game_in_">"Rejoindre la partie "</string>
     <string name="Game_Full">"Partie complète"</string>
-    <string name="US_East">"Amérique Centrale"</string>
-    <string name="US_West">"Amérique du Nord"</string>
+    <string name="US_East">"Etats-Unis de l'Est"</string>
+    <string name="US_West">"Etats-Unis de l'Ouest"</string>
```
Jeii20, line 62: The server was really changed its name, so the previous name was correct.

Jeii20 (changes requested):

res/values-fr/strings.xml
Teiko33: What do you mean by currently changed ? Because you can see that the translation is not correct, in english i just saw that servers are US east and west but it's still confused with America central and north which do not match at all 
Can you tell me a little more in detail ? ^^
And thanks for reply

Jeii20: **Shot 1:** English in-game
**Shot 2:** English strings file



49ba449c)

Jeii20: **Shot 1:** English in-game
**Shot 2:** English strings file

Teiko33: It changed on the update ? Because i see name in english image

And i just notice that it was Etat-Unis in french version image

Jeii20: Yes, you need to update the game to be able to see the changes.

Teiko33: It's not available on iOS but im gonna change it back 
Thanks

Jeii20 (approved):

res/values-fr/strings.xml
Teiko33: It was on the previous version 8.0.5 ( IOS ) on beta version but now it is fixed with the new update

PR #112, opened 2024-08-20
Update vietnam br language will be better

toitenxyzzz:
Let's add Vietnamese to

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -69,7 +69,7 @@
     <string name="Delay_Disconnect">"Atrasar a Desconexão"</string>
 
     <string name="Max_Players">"Máx. de Jogadores"</string>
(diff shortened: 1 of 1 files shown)


res/values-pt-rBR/strings.xml
```diff
@@ -69,7 +69,7 @@
     <string name="Delay_Disconnect">"Atrasar a Desconexão"</string>
 
     <string name="Max_Players">"Máx. de Jogadores"</string>
-    <string name="Min_Players">"Mín. de Jogadores"</string>
+    <string name="Min_Players">"Mín. de Jogadores. "</string>
```
Jeii20, line 72: You still keep changing the wrong language and keep inserting wrong periods. Don't forget you are editing the file `values-pt-rBR`, that is **Brazilian Portuguese**, and for **Vietnamese**, you must edit the `values-vi`!

PR #128, opened 2024-08-21
Update Indonesia Language (example)

Vynits:


Jeii20 (commented):
You don't need to base off completely on the mobile tutorial video, and that's what you did. The "example" is literally just example indeed.

And btw, in that string line, I don't think there is any dot needed, please leave it as it was.

You need to make accurate edits (such as correcting a text that you think it's wrong, inserting missing periods – **based on the English value** –, correcting any grammar issue, etc), that tutorial was actually to just help on the translation.

PR #133, opened 2024-08-21
Update values-pt-rPT

Jeii20:
Corrections for a better interpretation.

### DIFF
--- a/res/values-pt-rPT/strings.xml
+++ b/res/values-pt-rPT/strings.xml
@@ -165,7 +165,7 @@
     <string name="Blobicide">"Blobicídio"</string>
     <string name="Blobtrocity">"Blobtrocidade"</string>
     <string name="Blobaccre">"Blobacre"</string>
-    <string name="Blobastrophe">"Blobástrofo"</string>
+    <string name="Blobastrophe">"Blobástrofe"</string>
     <string name="Blobihilation">"Blobiquilação"</string>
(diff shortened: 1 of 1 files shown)


Jeii20:
I don't know what happened here, but please just consider the commit "Update values-pt-rPT".

res/values-pt-rPT/strings.xml
```diff
@@ -784,7 +784,7 @@
     <string name="Failed_to_send_mail_">"Falha ao enviar e-mail."</string>
     <string name="In_Game_Mail">"Mail no Jogo"</string>
     <string name="Nav_Buttons_">"Botões de Nav.:"</string>
-    <string name="account_warning">"Nunca compartilha a tua palavra-passe da conta ou informações pessoais com ninguém por qualquer motivo!"</string>
+    <string name="account_warning">"Nunca partilha a tua palavra-passe da conta ou informações pessoais com ninguém por qualquer motivo!"</string>
```
J0rgeF, line 787: "Nunca compartilhes a tua palavra passe..." sounds better

J0rgeF (changes requested):

res/values-pt-rPT/strings.xml
Jeii20: Oh, got it. I thought in Portugal you just use "partilhar" or something like that. Thanks for clarifying!

J0rgeF (approved):

PR #137, opened 2024-08-21
Update Indonesia Language (example)

Vynits:


Jeii20:
Could you please DM me on Discord? My user tag is @jeihehe, when you can, please send me something.

PR #142, opened 2024-08-21
Update strings. Bangla language (bn)

Nishan-cross:
Update strings. Bangla language (bn)


Jeii20 (approved):
Nice move. Thank you for understanding ^^

PR #145, opened 2024-08-22
Korean translation

dbswlstjr2:


chan084:
번역자 내놔

dbswlstjr2:
I already got the translator tag but I want to get more. Can I add it to another sub-account? My ID is 26463302

PR #146, opened 2024-08-22
Update strings.xml

RaRa40:
ID 14388435


TurkiAlanazi1:
> ID 14388435

You cannot translate languages ​​that are not your own.

PR #163, opened 2024-08-22
Update strings.xml si

Zombzs:


Zombzs:
Send me tag please on : 5551444

PR #164, opened 2024-08-22
Update values-ja (Japanese)

Jeii20:
### Urgent changes
1. パーマネント (Permanent): It's a correct term, but I think it's very literal by using katakana. According to all the researches I made, 永久に (/ēkyū ni/) is more appropriate in this case;
2. At string line 1030, the ジョイン (Join) was also very literal, so I had to change it to "traditional" literature (with use of kanjis + verb stablished);
3. At string line 929, it was been stablished the kanji "枠", that means "box", which doesn't make sense, so I changed to ブロブ (blob).

### Other changes
1. Changed the 永久に position from end to start, to meet with the Japanese structure order;
2. Translated more strings;
3. In コマンド (commands), I've fixed the double dot to ideographic literature;
4. Added a missing string compared to the `res/values`, it appears at line 1089 + already translated it to an approximate understandable way.
5. At line 930, it was been stablished a wrong translation. Since in English it is "ENABLED", in Japanese it was a "to enable" verb without any kind of tense, so I changed it to "ON" (turned **on**);
6. At line 931, it was been stablished a wrong translation, according to the English values. It wasn't really supposed to be a question (represented by "か？"), so I changed it to an accurate phrase + changed from "枠のカスタムカラー" (/waku no kasutamu karā/) to "ブロブのカスタム色" (/burobu no kasutamu iro/) to stablish an accurate reading.


Jeii20:
I identified out a missing string. `level_color_instructions` was been added below the `LEVEL_COLOR` in this recent change. I followed as it says on **Tips**.

Jeii20:
I deleted file by mistake, I will make these changes all again when I got time.

PR #173, opened 2024-08-23
New Update Spanish (Latin America)

theoox:
ID Nebulous: 3119626
Discord User: theodxrius

Corrected errors: inconsistency with uppercase and lowercase letters, poorly written texts and incorrect grammar, mistyped letters, correction of punctuation in numbers, for example (1.000 incorrect, 1,000 correct), words in English that are used exactly the same in Latin America, such as the word “zombie”.

Translation of (mayhem) to spanish as “modo caos”, correction of the warning that appears when you invite a player who is not online, and other words that needed to be translated from English to Spanish, correction of the label (COMP BANNED), the section (Message private), etc.

Error in Latin American Spanish with the words (8x 32x 64x etc.) since the letter "x" should be at the beginning to convey (SPLIT BY 8", "SPLIT BY 32", "SPLIT BY 64), etc. But as it is written now, the entire phrase is incorrect, as it reads "SPLIT 8 BY".

### DIFF
--- a/res/values-b+es+r419/strings.xml
+++ b/res/values-b+es+r419/strings.xml
@@ -1,4 +1,4 @@
-﻿<resources xmlns:tools="http://schemas.android.com/tools">
+<resources xmlns:tools="http://schemas.android.com/tools">
     <string name="Players">"Jugadores"</string>
     <string name="Control_">"Control:"</string>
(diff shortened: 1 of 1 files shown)


Jeii20:
He told me that he is about to make a new change for the Latin American Spanish, so this PR was been closed to avoid confusion. ^^

PR #175, opened 2024-08-23
German Translation Improvements

kp-leben:
id: 11462375

I made a new one because I got so confused with the old one... (am sorry)


Jeii20:
> Please request review from a previous german translator.

Yo, Greg
I am not a native German speaker, but I made some researches and I can say there are inaccuracies on the 2 first changed string lines in this file.

"Wöche" actually means just "Week" while the English string it's wrote "Weekly", so "Wöchentlich" was already correct. The same applies to the "Täg". These both shouldn't be changed, until where I reviewed.

PR #177, opened 2024-08-23
Update strings.xml

pheonix54:


pheonix54:
Can I got the tag ?
Id: 5551444

PR #182, opened 2024-08-23
Location is the place (Place)

Tigmax7:
Location is the place, not the rank. (place)

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -32,7 +32,7 @@
     <string name="Content_rating_of_the_app_may_change_when_finding_groups_">"Classificação de conteúdo do aplicativo pode mudar quando encontrar grupos."</string>
 
     <string name="_Rank_">" Posição "</string>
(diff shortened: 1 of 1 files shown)


res/values-pt-rBR/strings.xml
```diff
@@ -32,7 +32,7 @@
     <string name="Content_rating_of_the_app_may_change_when_finding_groups_">"Classificação de conteúdo do aplicativo pode mudar quando encontrar grupos."</string>
 
     <string name="_Rank_">" Posição "</string>
-    <string name="_Alias_">" Apelido "</string>
+    <string name="_Place_">" Apelido "</string>
```
Jeii20, line 35: Please never change the original string name, just the text in blank.

Jeii20 (changes requested):

Tigmax7:
I already did another one check it out

في الجمعة، ٢٣ أغسطس ٢٠٢٤, ٤:٥٤ م Jeii20 ***@***.***> كتب:

PR #187, opened 2024-08-23
Update strings.xml

Kemo201:
ID:  17067435


TurkiAlanazi1:
Meaning of word mistake.

Correct same 
DRAW: تعادل
It should not be changed.

TurkiAlanazi1 (commented):
mistake

PR #195, opened 2024-08-24
Update strings.xml

ikop498:
ID : 21731089


TurkiAlanazi1 (commented):
mistake
<string name="Translator">يترجم</string>
It should not be accepted.

PR #211, opened 2024-08-24
Update strings.xml

1964m:
ID 3717465

### DIFF
--- a/res/values-ar/strings.xml
+++ b/res/values-ar/strings.xml
@@ -1326,7 +1326,7 @@ Commands:
     <string name="Allow_Recombine">السماح بإعادة التجميع</string>
     <string name="Defenders">المُدافِعون</string>
     <string name="Attackers">المُهاجِمون</string>
-    <string name="MEGA_SPLIT">64x قسمة كبير</string>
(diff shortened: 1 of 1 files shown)


res/values-ar/strings.xml
```diff
@@ -1326,7 +1326,7 @@ Commands:
     <string name="Allow_Recombine">السماح بإعادة التجميع</string>
     <string name="Defenders">المُدافِعون</string>
     <string name="Attackers">المُهاجِمون</string>
-    <string name="MEGA_SPLIT">64x قسمة كبير</string>
+    <string name="MEGA_SPLIT">قسمة كبير</string>
```
Jeii20, line 1329: Please do not remove that. It may not appear in the English string (blue), but "64x" remains in every other language.

Jeii20 (changes requested):

res/values-ar/strings.xml
TurkiAlanazi1: > Please do not remove that. It may not appear in the English string (blue), but "64x" remains in every other language.

exactly

PR #223, opened 2024-08-24
Update arabic ar

heman20:
It was completed

id:5432184


zeynpcna2 (commented):
Why u did write same word

PR #237, opened 2024-08-24
Update Arabic ar Language

MaRs001V:
This word has two meanings, which are science and flag. If it is made like this, it will become flag, not science 

ID: 12962277


MaRs001V:
ID: 12962277

PR #250, opened 2024-08-25
add "plasma gift received!" , "Custom skin rejected" , "Deleted" , "F…

chan084:
…PS" , "CON" , "LAT" translate

ID : 18613714


Jeii20 (commented):
None of these texts are supposed to be translated. I've already said to please **never add** new strings without consulting the English file located at `res/values` first.

chan084:
So which one should I translate?

2024년 8월 25일 (일) 오후 7:17, Jeii20 ***@***.***>님이 작성:

Jeii20:
> So which one should I translate?

You should just translate or correct the already existing strings only. To add a string, you must submit an issue at the tab **Issues**. ^^

PR #251, opened 2024-08-25
Update Sinhala si Language

MaRs001V:
I translated approximately more than 30 sentences into Sinhala  
ID: 12962277


MaRs001V:
ID: 12962277

PR #252, opened 2024-08-25
Update Spanish Language

MiniPatron1234:
Espero que les sirva mi ayuda.
Nebulous ID 31190218


zeynpcna2 (changes requested):
don't change admin commands

PR #258, opened 2024-08-25
Update Arabic ar Language

MaRs001V:
I also noticed this when I was reviewing the script. I forgot to edit them, so I edited them now ^^


MaRs001V:
I did not know that I could modify the script after it was published, but after research I was able to modify it, so the modification below has become complete. I cannot delete this modification.

MaRs001V:
I really tried to delete it permanently from the files in "Pull requests" but I couldn't

MaRs001V:
Thank you so much, I really appreciate this from you 🤍

PR #283, opened 2024-08-25
Update strings.xml

t4h8:
id: 26845070


TurkiAlanazi1 (commented):
no need to change that
Change meanings of words.

PR #293, opened 2024-08-26
Update strings.xml tagalog translation (example)

Ezekiel477:
there is something wrong with the string (example)


Arjay180:
Other change it

res/values-tl/strings.xml
Arjay180, line 1: that tagalog word is not correct, please keep 'sign in' in tagalog

PR #298, opened 2024-08-26
Korean translation

trrrrre12442:


trrrrre12442:
Nebulas ID: 26463302

trrrrre12442:
I would like to receive a tag. Is that possible? 
@grog444

PR #299, opened 2024-08-26
Update language si

marya51:
id:21825988


MaRs001V:
The translation is wrong. The meaning of this translation is “the bay.” not "Buy" 
"මිලදී" This is correct not "බොක්ක"
Maybe when she translated she wrote "a" instead of "u", I don’t know

PR #332, opened 2024-08-26
Update strings.xml

rakanx40:
id 6740218


MaRs001V:
The translation is wrong because மொழிபெயர் refers to the verb. The correct translation is மொழிபெயர்ப்பாளர் because it refers to the person who performs this action. Because this is a Tag, the verb cannot be a Tag.

PR #340, opened 2024-08-26
Edites idioma Catalan

ZheiNB:
Faltas de signos y arreglo de números

### DIFF
--- a/res/values-ca/strings.xml
+++ b/res/values-ca/strings.xml
@@ -478,8 +478,8 @@
     <string name="Gladiator">"Gladiador"</string>
     <string name="Challenger">"Desafiador"</string>
     <string name="Win_1_1v1_Arenas_">"Guanya una Arena."</string>
-    <string name="Win_10_Arenas_">"Guanya 1.000 Arenes."</string>
-    <string name="Win_100_Arenas_">"Guanya 1.000 Arenes."</string>
+    <string name="Win_10_Arenas_">"Guanya 10 Arenes."</string>
+    <string name="Win_100_Arenas_">"Guanya 100 Arenes."</string>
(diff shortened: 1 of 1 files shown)


res/values-ca/strings.xml
```diff
@@ -1042,7 +1042,7 @@
     <string name="Buy">Buy</string>
     <string name="Sell">Sell</string>
     <string name="community_skin_requirement">You must purchase Enable Custom Skins to use community skins.</string>
-    <string name="You_can_purchase_this_community_skin_for_">You can purchase this community skin for</string>
+    <string name="You_can_purchase_this_community_skin_for_">You can purchase this community skin for:</string>
```
Jeii20, line 1045: Here it isn't supposed to be double dot, but spacebar.

Jeii20 (changes requested):

ZheiNB:
Jei hello how are you? I didn't understand the second question, what should
I translate?

res/values-ca/strings.xml
ZheiNB: Hi jei, sorry I didn't understand the second thing, what should I translate?

Jeii20:
@ZheiNB if you're about to make change, I think a translation is required on your own, otherwise you shouldn't make any relevant change there

ZheiNB:
I understand, sorry then Jei, one question, were all the corrections wrong? Or just the one with the two points?

Jeii20:
> I understand, sorry then Jei, one question, were all the corrections wrong? Or just the one with the two points?

@ZheiNB those texts you modified were already translated, so there was not a problem. If you don't know translation for that, leave them for the native speakers instead. The same applies especially to every other language you don't know (in cases like this, it's better not modify language files at no way).

ZheiNB:
Gracias Jei por explicarme ✨

PR #344, opened 2024-08-26
Update strings.xml

t4h8:
id: 26845070


MaRs001V:
Regarding the placement of the “,” in the numbers, it is correct to facilitate reading, and you removed it and this is negative. The rest of the translations are correct and there is no problem. It is better to explain what you changed in the words and meanings because they are all correct and there is no problem, And you have some incorrect edits. I ask you to focus carefully and good luck to you

PR #346, opened 2024-08-26
Update strings.xml

0Eliva:
ID: 31584934


MaRs001V:
There are many spelling errors and incorrect ones, such as the word اسم that I wrote. In the Arabic language, the hamza is not added to this word, and the word الدخول is not added, the word المستوى is not added, and some other words that are incorrect and do not change the meaning at all. I hope you check this before editing and good luck to you

PR #348, opened 2024-08-26
Update Lenguaje Espanish

ZheiNB:
La (i) minúscula fue cambia a (I) mayúscula 

Greg Hi, can I have the translator tag please? I will continue contributing to Nebulous 🙏  mí id: 13420431

### DIFF
--- a/res/values-es/strings.xml
+++ b/res/values-es/strings.xml
@@ -1203,7 +1203,7 @@ Esta es una lista de cosas * PROHIBIDAS * en las skins personalizadas. Si compra
     <string name="Session_Stats">Estadísticas de Sesión</string>
     <string name="Time_">Tiempo </string>
     <string name="Show_Session_Stats">Mostrar estadísticas de Sesión</string>
-    <string name="VALENTINES_DAY">"DÍA DE SAN VALENTíN"</string>
+    <string name="VALENTINES_DAY">"DÍA DE SAN VALENTIN"</string>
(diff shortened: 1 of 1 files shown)


Jeii20 (commented):
I can see you didn't insert the acute accent on the changed letter I. Is that intentional or is it supposed to be like that? According to my knowledge about the Spanish language, I think "VALENTÍN" is correct. I can see you changed from small (incorrect) to capital (correct), but probably missed an acute. ^^

ZheiNB:
If it is true, I then looked it up and it can be written in both ways, but
they have an accent. I didn't know how to rule it out.
Jei When no errors are found in any of the strings for the entire Spanish
language, how long can I review again?
I keep reading all the strings from beginning to end and I don't see any
text errors and when I find this one I said to myself that the search was
worth it but the correction I made is wrong.

Jeii20:
> If it is true, I then looked it up and it can be written in both ways, but
> they have an accent. I didn't know how to rule it out.
> Jei When no errors are found in any of the strings for the entire Spanish
> language, how long can I review again?
> I keep reading all the strings from beginning to end and I don't see any
> text errors and when I find this one I said to myself that the search was
> worth it but the correction I made is wrong.

You don't need to check absolutely all languages at all. Especially, it's recommended that you only check your native language or those languages you (really) know enough to be able to propose changes. I understand you just want to help, but checking all languages and sending too manu pull requests for practically the same reason may result in a confusion. ^^

PR #349, opened 2024-08-27
Update strings.xml

yosef3778:
lD: 29691360


MaRs001V:
The translation before your edit is correct. Please explain what is wrong with the word

PR #350, opened 2024-08-27
Update strings.xml

m7md10-ctrl:
ID:  25690087


MaRs001V:
The correct translation is مُترجم, not يترجم with the letter م, not ي, when writing the letter ي instead of م, this becomes a verb, and there is no verb, it becomes a Tag. The Tag is given to the person who does the action, so مُترجم is the correct and most accurate translation.

TurkiAlanazi1 (commented):
mistake. It should not be changed.

Previously correct
Translator - مترجم ✔️
he translates - يترجم ✖️

PR #351, opened 2024-08-27
Update strings.xml

mhde333:
ID: 23850441


Jeii20:
PR closed. Invalid changes. Please never modify the place you did right now (leaving a line break at incorrect place), and create a new PR with the valid changes.

PR #352, opened 2024-08-27
Update strings.xml

Hade-23:
id:27834320
I want the translators title


Hade-23:
What does closed order mean?

PR #357, opened 2024-08-27
Update strings.xml

Abdulazizgjb:
iD: 23152034


Jeii20:
PR closed. Please never change the original string names, because if changed and applied, they may result into an internal conflict.

PR #363, opened 2024-08-27
Update strings.xml (PT-BR)

AssaNebulous:
1707450

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -1364,7 +1364,7 @@ Comandos:
     <string name="Defenders">Defensores</string>
     <string name="Attackers">Atacantes</string>
     <string name="MEGA_SPLIT">MEGA SPLIT 64x</string>
-    <string name="Missing_Permissions">Permissões ausentes</string>
+    <string name="Missing_Permissions">Sem Permissões</string>
     <string name="DELETE_ACCOUNT">EXCLUIR CONTA</string>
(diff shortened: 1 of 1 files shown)


res/values-pt-rBR/strings.xml
```diff
@@ -1364,7 +1364,7 @@ Comandos:
     <string name="Defenders">Defensores</string>
     <string name="Attackers">Atacantes</string>
     <string name="MEGA_SPLIT">MEGA SPLIT 64x</string>
-    <string name="Missing_Permissions">Permissões ausentes</string>
+    <string name="Missing_Permissions">Sem Permissões</string>
```
Jeii20, line 1367: Original text was correct. The text you tried to correct wouldn't make sense if changed like that, because in English it would mean "No Permissions".

Jeii20 (changes requested):

res/values-pt-rBR/strings.xml
AssaNebulous: Entendi, so tentei alterar por conta que Permissões Ausentes não é tão utlizado

Jeii20: "Permissões em falta" would sound better in this case then.

AssaNebulous: Ok

Jeii20:
PR closed. Another PR with the change at same line is identified at #368.

PR #364, opened 2024-08-27
Update strings.xml

mhde333:
ID: 23850441

### DIFF
--- a/res/values-ar/strings.xml
+++ b/res/values-ar/strings.xml
@@ -120,7 +120,7 @@
     <string name="This_player_is_not_signed_in">"هذا اللاعب لم يسجل الدخول"</string>
 
     <string name="Online">"مُـتَـصِـل"</string>
-    <string name="Unknown">"غير معروف"</string>
+    <string name="Unknown>"  مخفي"</string>
     <string name="Offline">"غـيـر مُـتَـصِـل"</string>
     <string name="Achievement_Earned">"تم تحقيق الإنجاز"</string>
     <string name="REFRESH">"تحديث"</string>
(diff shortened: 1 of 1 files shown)


res/values-ar/strings.xml
```diff
@@ -120,7 +120,7 @@
     <string name="This_player_is_not_signed_in">"هذا اللاعب لم يسجل الدخول"</string>
 
     <string name="Online">"مُـتَـصِـل"</string>
-    <string name="Unknown">"غير معروف"</string>
+    <string name="Unknown>"  مخفي"</string>
```
Jeii20, line 123: Please return the blue " at "Unknown", and remove the huge space there.

Jeii20 (changes requested):

res/values-ar/strings.xml
MaRs001V: > Please return the blue " at "Unknown", and remove the huge space there.

Unfortunately, the translation is wrong. 
" مخفي" "It means" hidden 
"UnKnown" It means "غير معروف" Or "مجهول" Both are correct 
So leaving it as it is, is the most appropriate translation

Jeii20: @MaRs001V I am talking about this:

Jeii20: You switched the " and the >, and it may cause internal conflict if you do it. Please return it back.

TurkiAlanazi1: Wrong translation 

hidden>مخفي<
Unknown>غير معروف<
It should not be accepted. 🔴

PR #372, opened 2024-08-27
Update strings.xml

Hade-23:
id:27834320
I want the translator's title


Jeii20 (changes requested):
Why did you return it into English?

Hade-23:
This is the solution as far as I know.

Jeii20:
PR closed. As it says in **Tips**:

"If you are revising something that is already translated, an explanation is required".

PR #373, opened 2024-08-27
Update strings.xml

Abdulazizgjb:
id 23152034


Jeii20:
PR closed. Please never change the `<item>` or the `</item>`, because if changed and applied, it may result into internal conflict.

PR #377, opened 2024-08-27
Update Arab language

Abdulazizgjb:
id: 23152034


MaRs001V:
The translation is wrong

Jeii20:
PR closed. You modified the first line which isn't allowed, and removed the `<string array>` at line 1414. Please create another pull request with the valid changes – never modify the lines you did right now.

PR #383, opened 2024-08-27
Update Arab language

Abdulazizgjb:
Correction of some incorrect words


Jeii20:
PR closed. Wrong change at line 1214. Please never remove the string name, otherwise if changed or applied, it may result into an internal conflict.

PR #385, opened 2024-08-27
Update strings.xml

Abdulazizgjb:
Correction of some incorrect words

id: 23152034


Jeii20:
As it says on **Tips**:

"Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is required".

Please provide it here

PR #392, opened 2024-08-28
Update strings.xml

Yousab001:
ID:23868983


Jeii20:
As it says on **Tips**:

Majaaaa1:
@Yousab001 This is not valid translation,please the next time you work on it, make sure it's done nicely. Do not make changes except where is necessary, be careful with the strings.

PR #401, opened 2024-08-28
Update strings.xml

TurkiAlanazi1:
improvements


TurkiAlanazi1 (commented):
now applies Message with skins, hats, pets

PR #402, opened 2024-08-28
[5032837] Correct Grammar & Improve the "Improvements"

0x8fv:
I went back and changed a lot of the last commit since the translations were very very bad and felt very Google translated.


res/values-de/strings.xml
```diff
@@ -9,8 +9,8 @@
     <string name="Button_Margin_">"Schaltflächen Position:"</string>
     <string name="Buttons_">"Schaltflächen:"</string>
     <string name="DEFAULTS">"Vorgaben"</string>
-    <string name="WEEKLY">"Wöchentlich"</string>
-    <string name="DAILY">"Täglich"</string>
+    <string name="WEEKLY">"wöchentlich"</string>
+    <string name="DAILY">"täglich"</string>
```
Jeii20, line 13: Lines 12 and 13: Why removed the capitals? Btw, this was actually supposed to be 100% capirals like in `res/values`.

Jeii20 (changes requested):

res/values-de/strings.xml
0x8fv: It was lowercase before too, I see in other language xml files it's always uppercase. It doesn't really matter, though, because those are the translations for the leaderboard, which is always set to all uppercase by the game.

I reverted those changes since those words aren't "Nomen" and should be lowercase for German sentences.

Jeii20: In Android, the fact that it's uppercase is true, but you must think about the iOS users too.

0x8fv: I updated the commit, it's now uppercase

Jeii20 (approved):
Thanks for your time ^^
Approved the changes, can merge into main if possible.

0x8fv:
@Jeii20 Can you set this merge request to approve?

PR #403, opened 2024-08-28
Update strings.xml

Marennnnn:
ID: 27391921


Jeii20:
As it says on **Tips**:

MaRs001V:
You really have a lot of mistakes

PR #405, opened 2024-08-28
Update strings.xml

Kemo201:
ID: 17067435


Jeii20:
As it says on **Tips**:

PR #412, opened 2024-08-28
Update strings.xml

305h:
ID: 19011344


Jeii20:
PR closed. Wrong changes at lines 951 and 1053. Please never change "true" next time, because it may cause internal conflict.

PR #415, opened 2024-08-28
Turkish Translate

zeynpcna2:
added PLAYER NOT FOUND

### DIFF
--- a/res/values-tr/strings.xml
+++ b/res/values-tr/strings.xml
@@ -426,10 +426,10 @@
     <string name="Purchase_Successful_">"Satın alma işlemi başarı ile tamamlandı."</string>
     <string name="You_may_need_to_reconnect_to_use_the_purchase_">"Ürünü, kullanabilmeniz için oyuna yeniden bağlanmanız gerekebilmektedir."</string>
     <string name="Confirm_Purchase">"İşlemi Onaylayın"</string>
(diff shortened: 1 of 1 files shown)


res/values-tr/strings.xml
```diff
@@ -1389,6 +1389,7 @@ Oturum açmak istemiyorsanız, tek oyunculu modumuzu oynamaya devam edebilirsini
     <string name="Account_must_be_1_day_old_to_send_mail_">Posta gönderebilmeniz için hesabınızın en az 1 günlük olması gerekmektedir.</string>
     <string name="Translator">Çevirmen</string>
     <string name="SUCCESS">BAŞARI</string>
+    <string name="PLAYER NOT FOUND">"OYUNCU BULUNAMADI"</string>
```
Jeii20, line 1392: It must contain the _, like: `PLAYER_NOT_FOUND`. Also, you need to either submit an [**issue**](https://github.com/simplicialsoftware/nebulous-translations/issues) if this string name doesn't exist in any of the files, or look yourself for the string at `res/values` (English file) and try adding at the correct line if it appears, so it may work well in-game, and it usually contains `translatable="false"`, as it says on **Tips**. ^^

Jeii20 (changes requested):

PR #420, opened 2024-08-29
Update strings.xml change Romain

AlExX2457:
update string  change romain

### DIFF
--- a/res/values-ro/strings.xml
+++ b/res/values-ro/strings.xml
@@ -36,7 +36,7 @@
     <string name="_Score_">" Scor "</string>
     <string name="Score__">"Scor: "</string>
     <string name="Recombine_">"Recombină:  "</string>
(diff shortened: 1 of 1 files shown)


res/values-ro/strings.xml
```diff
@@ -128,7 +128,7 @@
     <string name="Friend_Request_Sent">"Cerere de prietenie trimisă"</string>
     <string name="ERROR">"EROARE"</string>
     <string name="CONTINUE">"CONTINUĂ"</string>
-    <string name="Global_" formatted="false">"Global%"</string>
+    <string name="Global_" formatted="false">"Global formatata fals"</string>
```
Jeii20, line 131: You shouldn't translate and provide the `formatted=false`. "Global" seems better here.

Jeii20 (changes requested):

Jeii20:
PR closed. Same change at #421

PR #421, opened 2024-08-29
Update romain ro (exemple)

AlExX2457:
They forgot to translate both words completely (exemple)

### DIFF
--- a/res/values-ro/strings.xml
+++ b/res/values-ro/strings.xml
@@ -36,7 +36,7 @@
     <string name="_Score_">" Scor "</string>
     <string name="Score__">"Scor: "</string>
     <string name="Recombine_">"Recombină:  "</string>
-    <string name="WINNER_">"C ŞTIGĂTOR"</string>
+    <string name="WINNER_">"CÂŞTIGĂTOR"</string>
     <string name="IDLE">"INACTIV"</string>
     <string name="Name_Invalid_">"Nume invalid."</string>
     <string name="Name_Taken_">"Nume luat."</string>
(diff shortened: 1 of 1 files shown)


res/values-ro/strings.xml
```diff
@@ -128,7 +128,7 @@
     <string name="Friend_Request_Sent">"Cerere de prietenie trimisă"</string>
     <string name="ERROR">"EROARE"</string>
     <string name="CONTINUE">"CONTINUĂ"</string>
-    <string name="Global_" formatted="false">"Global%"</string>
+    <string name="Global_" formatted="false">"Global formatat fals"</string>
```
Jeii20, line 131: Wrong change here. You shouldn't translate and provide the `formatted="false"`.

Jeii20 (changes requested):

res/values-ro/strings.xml
AlExX2457: Alright

AlExX2457: So is not right, why, that said there Without%

AlExX2457: There up at winner is Câstigator not how they translate there, can i get my tag translator or i need to try again and to see better there if is wrong or no something

Jeii20: You just need to correct the changes

AlExX2457: Alright, i changed right now

AlExX2457:
Can you check now @Jeii20 (Thank you in advance)

Jeii20:
PR closed. Another PR with requested change identified at #480.

Jeii20 (commented):
.

PR #423, opened 2024-08-29
Update strings Portuguese (Brazil)

Aybar1881:
Translated Success, "BATALHA REAL" has been changed to "BATTLE ROYALE" cuz its special name. Translated Mod names.


Jeii20:
PR closed. Too many wrong changes.

Aybar1881:
> PR closed. Too many wrong changes. Plus: you actually edited Portugal file. "Success" translation is correct, but the other rest are invalid changes

So can i just Change for Succes? And open New PR for success?

Jeii20:
You can look for other texts. These you tried to change must remain the same, especially "MEGA SPLIT 64X" and "CRAZY SPLIT 32X", because of the typeface that may not work well.

Aybar1881:
> You can look for other texts. These you tried to change must remain the same, especially "MEGA SPLIT 64X" and "CRAZY SPLIT 32X", because of the typeface that may not work well.

Got it thanks😊

PR #439, opened 2024-08-29
Update strings.xml ar Arabic

lempo1:
my id:4889432

Correct words


zeynpcna2 (commented):
No change

lempo1:
> No change

In the Arabic language, the tanween, hamza, and other vowels are placed on the letters, which are called in the Arabic language, so that they can be read correctly.

PR #441, opened 2024-08-29
Update strings.xml

Ghost98609:
ID: 30099340


Jeii20:
PR closed. Wrong change at line 462. Please never change the original string name.

Majaaaa1:
@Ghost98609 hi
There is no need to open a ticket when everything necessary has already been said here.Please pay attention to the instructions and correct the translation where necessary and do not touch the original part of the file, only the part intended for translation.Take your time and check the complete translation, I understand that you want a tag, but that is not the reason to give a tag for incorrect changes.

PR #452, opened 2024-08-30
Update strings.xml

ok793n:
ID :9582073


Jeii20:
As it says on **Tips**:

MaRs001V:
The first sentence is the most accurate because adding the fatha to the sentence makes it in the past

PR #463, opened 2024-08-30
Update strings.xml

Bek0o0o:
You call people below who are not 18 years old ´´Du´´ and it is pretty disrespectful calling  adults ´´Du´´ instead of ´´Sie´´.  In other games (even for children) are translated like that. I hope you understand my explanation and agree with my decision.

ID: 10072367
Discord: bek0o0


0x8fv:
>  is pretty disrespectful
Actually, it's not. It's just the formal variety.

PR #466, opened 2024-08-30
Update strings.xml

Bek0o0o:
Schub is the correct translation for boost.

ID: 10072367
Discord: bek0o0


Jeii20:
PR closed. Invalid changes at lines 398, 439, 907, 1020, 1021, 1025, 1177, 1179, 1198, 1250, 1251, 1252, 1253, 1254 and 1327. Please **never** change the original string name (in blue), otherwise it may result in internal conflicts.

PR #487, opened 2024-08-31
Update strings.xml

zelus777:
ID: 25893191


Jeii20:
PR closed. Invalid change at line 1352. Please **never** change the original string name (in blue), otherwise it may cause internal conflict.

PR #502, opened 2024-08-31
Update strings.xml

zelus777:
I put diacritics for the words because they differ in pronunciation without diacritics and also with diacritics. The one I put in is correct this is an example of correcting words from wrong to correct the formation of words is differentiated by pronunciation 
ID: 25893191


Jeii20 (approved):

zelus777:
Will I get the title of translator?

في السبت، ٣١ أغسطس ٢٠٢٤ ٧:٥٣ م Jeii20 ***@***.***> كتب:

PR #527, opened 2024-09-01
Update Estonio

LaFurraDiana:
Id nebulous:14282977

;b


AnkaKoxu:
2 words should start here with capital letters to match with english strings.

PR #543, opened 2024-09-02
Update strings.xml

Mira14150:
ID: 27391921


CritiKangel (changes requested):
Changed to a more standarised use.

CritiKangel (approved):
Approved change in typography.

PR #548, opened 2024-09-03
Portuguse BR Translation Improvements

santos07x:
id:24332615


Jeii20:
PR closed. Please keep here as "PRIV.", because it's already understandable for players

PR #551, opened 2024-09-03
Update strings.xml Ro(Romain)

AlExX2457:
Update Strings.xml Ro
They're wrong every time, so many times I keep saying it's spelled wrong !
My ID is: 20713867

The punctuation marks are wrong, that's not how we are used in Romanian, totally wrong (most of them)


AssaNebulous (commented):
hello, in some changes you removed the punctuation from the end, I think these changes are incorrect.

I didn't find anything that confirms your request in saying that Romanian does not use punctuation at the end of sentences.

AnkaKoxu:
Could you explain why full stops are not used in the majority, give any examples?

why is there a dot in some sentences? e.g. here?

AlExX2457:
They left it like that, most of them are without punctuation marks, only with a comma, when we finish a word we don't put (.)
Some I forgot to remove, that's why it's like that, I missed many
Sometimes we use punctuation marks at the end of sentences

Jeii20:
> Hello There!
> They left it like that, most of them are without punctuation marks, only with a comma, when we finish a word we don't put (.) just sometimes some stories have long sentences
> 
> Some I forgot to remove, that's why it's like that, I missed many

In-game it's needed that they keep like it were before @AlExX2457

AlExX2457:
I understand but this seems wrong to me 
@Jeii20

Jeii20:
First please fix the punctuations you removed

AlExX2457:
Ok, i will fix it, but you promis you will accept my translator into romain?

AlExX2457:
@Jeii20 Please check if it's ok now! I want to be a Romanian translator one day! Nothing else! Please, how long do I have to wait?

AnkaKoxu:
It has become such a chaos with these punctuations that it is not clear what is correct. Match everything with english' strings. Because once you give those dots and once you don't, and once change them. It is no longer clear whether this is even a good thing, as you have only 35 improved latest out of over 1000+.

AlExX2457:
I deleted the tab with the power points as Jai told me, the rest is fine, I translated what was not translated

AssaNebulous:
You deleted all the code...

AlExX2457:
No, i didn't,go look up, I only deleted what he said

Jeii20:
PR closed. Commit file has been deleted (probably by mistake).

PR #553, opened 2024-09-03
Update Portuguese br language (example)

Lilie19:
They missed a dot in a string


Jeii20:
PR closed. Please never modify the formatting `%s`, `%d`, `\n` or similar. Also, you don't need to inspire completely on the tutorial video, as it says, it's just an **example**.

PR #559, opened 2024-09-03
id=20781055

Taki3600:
Translating some words and correcting errors

### DIFF
--- a/res/values-ar/strings.xml
+++ b/res/values-ar/strings.xml
@@ -336,7 +336,7 @@
     <string name="_in_a_Teams_Game">" في لعبة الفرق"</string>
     <string name="_dots">" النقاط"</string>
     <string name="_player_blobs">" دائرة اللاعب"</string>
-    <string name="Absorb_">"استوعب "</string>
+    <string name="Absorb_">"استيعاب "</string>
     <string name="Collect_">"جمع "</string>
 
     <string name="FFA_CLASSIC">"حر للجميع الكلاسيكية"</string>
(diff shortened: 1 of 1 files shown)


res/values-ar/strings.xml
```diff
@@ -1211,15 +1211,15 @@
...
+    <string name="GIVE_PLASMA">"اهداء بلازما"</string>
     <string name="Specify_amount_">حدد المبلغ:</string>
     <string name="Specify_amount_again_to_confirm_">حدد المبلغ مرة اخرى للإكمال:</string>
     <string name="Amounts_did_not_match_">المبلغ غير مُتطابق!</string>
     <string name="RECEIVE_PLASMA">تلقي البلازما</string>
     <string name="MULTIBLOB">مُتعددة النُسَخ</string>
     <string name="Play_Nebulous_For_1_Year_">ألعب نيبلوس لمُدَّة 1 سنة!</string>
     <string name="Play_Nebulous_For_X_Years_">العب نيبلوس لمُدَّة %d سنة!</string>
-    <string name="Starting___">بدء…</string>
+    <string name="Starting___">بدأ…</string>
```
mody515, line 1222: The edit he made is wrong, the old translation is correct

res/values-ar/strings.xml
```diff
@@ -1418,7 +1418,7 @@ Commands:
         <item>"صغير"</item>
         <item>"عادي"</item>
         <item>"كبير"</item>
-        <item>"ضخم"</item>
+        <item>"كبير جدا"</item>
```
mody515, line 1421: Also the same problem here, the old translation is the correct one

res/values-ar/strings.xml
Taki3600: Can you explain to me more where the error is? 
If you are correct, I will delete this translation 
If you're taking this out of revenge, 
**stop bothering us**

res/values-ar/strings.xml
Taki3600: I want answer this because it's clear that you just want to bother me 🤦🏻
I will not ignore your criticisms, if they are correct, I will correct the errors 
Ty

Taki3600: Okay, I corrected the mistakes

res/values-ar/strings.xml
Taki3600: I fixed it

Jeii20 (approved):
Already approved, can merge into main as soon as possible.

PR #563, opened 2024-09-04
strings.xml 업데이트

Limsoohyeon:


Cedricnebu:
I dont think its needed

Jeii20:
PR closed. Please never modify the original string name (in blue).

PR #569, opened 2024-09-04
Update strings.xml

zeroo313:
ID: 31253902


Jeii20:
PR closed. Invalid changes at all lines. Reasons:

Line 1357: You don't need to base off completely on the English texts in all string lines. For that, you can check `res/values` (English file).
Line 1358: If the punctuation was inserted, you don't need to remove.
Line 1359: It literally says "Check your connection". As of line 1357, please check `res/values` to make valid change.

zeroo313:
Now what should I do? I have changed it more than once

PR #586, opened 2024-09-05
Update portuguese br language

Enzinhozs:
They missed a dot in a string (pérolas)


Jeii20:
PR closed. You don't need to inspire completely by the tutorial video + it already says in my pinned comment. You removed punctuation on where was required that it stays here.

PR #587, opened 2024-09-05
Update English Language

Captain7816:
They missed a question mark on a string.
And this is my Nebulous's ID for translator Tag : 22539568


Jeii20:
PR closed. Invalid change. Please never modify the original string name, otherwise it ain't gonna work as it should. The string names need to always follow the `res/values` (English file). Please submit a new PR with valid changes.

PR #588, opened 2024-09-05
Update french fr language

TigerNebu:
They missed in some strings.
My id in case : 13862477

### DIFF
--- a/res/values-fr/strings.xml
+++ b/res/values-fr/strings.xml
@@ -21,7 +21,7 @@
     <string name="Avatar_">"Skin :"</string>
     <string name="Name_">"Nom :"</string>
 
-    <string name="Loading___">"Chargements… "</string>
+    <string name="Loading___">"Chargement… "</string>
     <string name="Connecting___">"Connexion… "</string>
 
     <string name="Are_You_Sure_">"Êtes-vous sûr ?"</string>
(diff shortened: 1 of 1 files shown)


res/values-fr/strings.xml
```diff
@@ -58,13 +58,13 @@
     <string name="SERVER">"SERVEUR"</string>
     <string name="Joining_game_in_">"Rejoindre la partie "</string>
     <string name="Game_Full">"Partie complète"</string>
-    <string name="US_East">"Amérique Centrale"</string>
-    <string name="US_West">"Amérique du Nord"</string>
+    <string name="US_East">"Amérique de l'ouest"</string>
+    <string name="US_West">"Amérique de l'est"</string>
```
Jeii20, line 62: The server names at lines 61 and 62 must not be redone back to previous names, as they were also changed in English. Proof is at English file `res/values` and in-game English. Please return the original names here

Jeii20 (changes requested):

res/values-fr/strings.xml
TigerNebu: What do i have to do now sir ?

Jeii20: You can edit the file through this same pull request (consequently submitting another commit here) and return the names as they were

TigerNebu: Oh okay i can also add some errors or i need to make a new request ? And maybe find others dots, because i didn’t checked everything tbh. And i want the translator tag but i don’t know how many words i need to translate right.

Jeii20: You just don't need to fix every line, but those you think that are really in need. You don't need to submit another PR (Pull Request), as I already said, you can just access this PR, then submit another commit here with the fixes

TigerNebu: Okay well thank you i’ll do it as soon as i can.

PR #589, opened 2024-09-05
Corrected an English form

Captain7816:
There was an error in the wording of the sentence. My Nebulous's ID : 22539568


Jeii20:
PR closed. Please never modify the original string name. It needs to keep the same as it is on English file: `res/values`, otherwise it ain't gonna work as it should.

PR #610, opened 2024-09-06
Update strings.xml

21499465:
ID: 18772929


Jeii20:
Please provide a reason for this change, as it says on **Tips** topic:

""Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is required."

PR #646, opened 2024-09-08
Update strings.xml

hamtoo77:
id: 12003952


Jeii20:
PR closed. Please never change the original string name (you did it 3 times), because it ain't gonna work as it should if modified.

PR #652, opened 2024-09-08
Update strings.xml

zelus777:
ID: 25893191


Jeii20:
PR closed. Unnecessary change that results in the same.

PR #660, opened 2024-09-09
[5032837] Revert bad translations

0x8fv:
The word "Boost" is actually German. We in Germany use the word for many things, there are many English words that were applied to the German vocabulary (we call English words that became German "Eingedeutscht").
The word "Boost" would fit much better and is grammatically correct, unlike "Schub".


0x8fv:
> I would just request to keep the "B" as capital letter before approval.

Done.

Jeii20 (approved):
Thanks for your time ^^

PR #664, opened 2024-09-09
Update Albania

VV-2:
There were a lot of words, the beginning of their letters in the small form. So I modified it

ID: 1644067


Jeii20 (approved):

PR #668, opened 2024-09-09
Update strings.xml admin inventures

Rlatjwjddazz:
I translated !allow into “허용하다 명렁어” 
As such, I translated English into Korean about the room management authority 

My id 24409441
Thanks

### DIFF
--- a/res/values-ko/strings.xml
+++ b/res/values-ko/strings.xml
@@ -1311,17 +1311,17 @@ Nebulous 규칙\n
     <string name="The_winner_will_receive_x_plasma_">승자는 %s개의 플라즈마를 획득합니다!</string>
     <string name="admin_commands">"
 명령어:
-!kick [0–31]
-!reset
-!welcome {message}
-!difficulty [0-3] or [1-4] for Campaign
-!hidden
-!guests
-!slowchat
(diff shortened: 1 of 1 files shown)


res/values-ko/strings.xml
```diff
@@ -1311,17 +1311,17 @@ Nebulous 규칙\n
...
-!welcome {message}
-!difficulty [0-3] or [1-4] for Campaign
-!hidden
-!guests
-!slowchat
-!pausechat
-!mute {id}
-!admin {id}
-!allow {id}
+!kick [0–31] [0-31] 추방 명령어
```
Jeii20, line 1314: Why repeated the "[0-31]"?

Jeii20 (changes requested):

res/values-ko/strings.xml
AnkaKoxu: Tip from me also, to explain **every** command. !allow reset and !winner has been abandoned.

Rlatjwjddazz:
So, is there no language that I can translate into?

2024년 9월 9일 (월) 오후 11:24, Ania ***@***.***>님이 작성:

PR #671, opened 2024-09-09
Update strings.xml

nsi117:
ID: 4374272


Jeii20:
PR closed. You changed the original string name at line 40. Please never do it.

PR #673, opened 2024-09-09
Update strings.xml

Hade-23:
I have corrected some wrong words or extra vowels. id:230455657


Hade-23:
When can I get the translator title?

PR #678, opened 2024-09-09
Update strings.xml

305h:
ID:31311113


AnkaKoxu:
Please, ensure unity of translation.

Why you translated MAIL as POST, but you left it afterwards like in english?

"WRITE_MAIL">"SCRÍBHINN **MAIL**"

Jeii20 (changes requested):
I agree with @AnkaKoxu.
Please either replace "Post" in strings that contain "Mail" or leave all of them untranslated.

PR #684, opened 2024-09-10
Atualizar Português BR-PT ID:26702913

Yutinha235:
correções dos modos Ice Split e Pop Split.

Dc:yuta.lz

Thanks bye.


Jeii20:
They aren't really modes but notifications. I don't think these are needed, since people are already used to the current texts. I will be closing this. Next time, please consider the custom of people in-game, as number of brazilian-portuguese-speaker players is great.

PR #687, opened 2024-09-10
Update strings.xml

Moayadneb:
ID 246213


TurkiAlanazi1:
> ID 246213

hi you can put space between words

Moayadneb:
> > ID 246213
> 
> hi you can put space between words

Done Thanks for telling me that

PR #691, opened 2024-09-10
Update Turkish language strings.xml

Nebuksk:
name="Show_skins">"Show Skins"</string>

Verified Account ID: 29637519


Majaaaa1:
@Nebuksk Please be patient, it has to go through the approval process. It should not be accepted, only one translated word, in order to give a tag, please show some effort and correct if something else is needed.
I will close the ticket, please pay attention to future requests here.

Jeii20:
Unfortunately, your PR will be closed by the reason: Please never change the blue colour text like you did, otherwise it ain't gonna work as it should. Please only change the white text (in dark mode) or black text (in light mode).

PR #693, opened 2024-09-10
valores-pt-rBR// Melhoria na linguagem pt-rBR

Banned44nebu:
Melhoria na linguagem pt-rBR
Concerto de tradução na linha "1316"
"Nebulosas" para "Nebulous".

ID NEBULOUS : 23262964


Jeii20:
PR closed. You just changed to informal tone, which can negatively affect comprehension, so changes weren't really needed. However, you made a very good attempt this time, congratulations for that.

PR #695, opened 2024-09-10
Update Árabe عربي Language (example)

Milk5999:


Jeii20:
PR closed. Unnecessary change at both string name and at format. Please, before making any change, always base on the English file: `res/values`

PR #702, opened 2024-09-11
Update strings.xml

Hade-23:
I have edited some wrong words to correct. 
id:230455657


Hade-23:
Will I get the title of translators?

Jeii20:
@Hade-23 
You need to provide the reason why the previous texts are wrong, as it says on **Tips**:

Hade-23:
@Jeii20 
I have modified some words to completely change the meaning of the word such as: البيوض It should be
البيض Another example الملف الشخصي في Google
And it should be الملف الشخصي على Google
Do you think this does not change the meaning of the words? For everyone to understand

PR #703, opened 2024-09-11
Update strings.xml

zelus777:
ID: 25893191


zelus777:
> ID: 25893191
I corrected the word “managers” from correct to more correct. I hope you understand what I mean

zelus777:
I corrected the word “managers” from correct to more correct. I hope you understand what I mean

Jeii20:
@zelus777 
If the previous text was correct, I don't think such change was needed.

PR #706, opened 2024-09-11
Update portuguese br language

JhonnySzs:
They missed a dot in a string

My ID: 11766037

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -303,7 +303,7 @@
     <string name="Survival_Domination">"Dominador de Sobrevivência"</string>
     <string name="I_Will_Survive">"Eu vou sobreviver"</string>
     <string name="Soccer_Domination">"Dominador do Futebol"</string>
-    <string name="Soccer_Hat_Trick">"Hat-Trick do Futebol"</string>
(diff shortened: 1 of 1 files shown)


Jeii20:
PR closed. Invalid changes in too many lines.

### Plus
- Please never modify the blue colour text, such as you did at lines 450, 697, 700, 895, 984 and 985, otherwise they ain't gonna work as it should. They need to remain the same as of `res/values` (English file).
- Please never remove the formattings, such as `%d`, `%s`, `\n` or similars like you did at lines 474, 654, 837, 984 and 985. They need to keep there.
- Avoid making too literal translations, such as you made at line 306. "Hat-trick" is greatly recognised by brazilian-portuguese-speakers.
- Please never remove the capital letters if they are kept in previous texts, like you did at lines 697 and 793.
- Please don't make the string length too long, such as you did at line 785. If they are abbreviated, it's for a good reason.
- Do not try to redo already recently changed texts or the already correct texts, like you did at lines 456 and 560.
- You don't need to inspire completely by the tutorial video (and I am talking about the title and description you provided for this PR).

Please take them all as consideration for next time. Never do "random" changes just for the translator tag, you must do valid changes always providing reason for them. Thanks ^^

PR #709, opened 2024-09-11
Update portuguese br

JhonnySzs:
Words are missing

My ID: 11766037


Jeii20:
PR closed. Please keep it as it was by 2 reasons: it was recently changed and it doesn't make sense returning back to how it was, and because it's based on English file (`res/values`).

PR #713, opened 2024-09-11
Update strings.xml

fareees1:
ID:24343417


res/values-tr/strings.xml
```diff
@@ -46,7 +46,7 @@
     <string name="PLAYING">"OYNANIYOR"</string>
     <string name="MULTIPLAYER">"ÇOK OYUNCULU"</string>
 
-    <string name="JOIN">"KATIL"</string>
+    <string name="JOIN">"KATıLMAK"</string>
```
Jeii20, line 49: Why isn't the "I" capital here?

Jeii20 (changes requested):

PR #730, opened 2024-09-12
Update strings.xml

Hade-23:
It was in English and I made it in Spanish so everyone could understand.id:4374272


Jeii20:
PR closed. Invalid change. "Piel" is too literal and means human skin. Please keep it as "skin" because it's greatly accepted on in-game context.

PR #738, opened 2024-09-12
Update strings.xml

05a7:
id: 27278361


05a7:
إستيعاب✅
استيعاب❎

Jeii20:
Please provide a reason for this change and why the previous text was wrong, as it says on **Tips**:

PR #758, opened 2024-09-14
valores-pt-rBR

Banned44nebu:


Jeii20:
PR closed. Make sure you are editing the PORTUGAL version.

- "equipes" is not correct (the correct is "equipas").
- "Restaurar" is not correct (the correct is "Redefinir").
- "escondida" is not correct (the correct is "oculta").
- "permissãoes" has an incorrect grammar (typo).
- "Vencedor com partida de tempo" can negatively affect the interpretation.

PR #759, opened 2024-09-14
Update strings.xml

Banned44nebu:


Jeii20:
PR closed. This change is not really correct. Here is a few information:

**In dutch, "bericht welkom" is an incorrect expression for the meaning of "welcome message", and the actually correct expression is "welkomstbericht".**

If you aren't sure of a language that's not your native, please **never** manage it.

PR #765, opened 2024-09-15
Update strings.xml

fareees1:
ID :24343417


Jeii20:
PR closed. According to a feedback, you keep changing the same text that was already been told several times that the previous texts were already correct. Next time, please provide a reason for these changes with valid arguments.

PR #766, opened 2024-09-15
Update README.md

Jeii20:
### Changes
- This change includes 3 new topics at the **Tips**, to clarify what is to do and what is not to do, based on the tutorial video.


Jeii20:
I think the last topic should appear at the end of file (before License), but you can decide about it.

PR #771, opened 2024-09-16
Translation into Spanish

Wuzzkah:
It's not such a serious mistake, but the correct way is "elige" not "elije".  "Elije" is not a valid word in Spanish. Tere are conjugations with the letter, but not in this case.


Jeii20 (approved):

PR #773, opened 2024-09-16
Update strings.xml

z4mw:
27278361


Jeii20 (changes requested):
It's not really supposed to have an exclamation mark.

PR #797, opened 2024-09-19
Update strings.xml

Banned44nebu:


Jeii20:
PR closed. Wrong changes at lines 1343 and 1347.

## Reasons

1. Translation for "kick" is not "remover", but "expulsar";
2. Translation for "hidden" needs to keep as "escondida", because it's used everywhere in brazilian version. If it was for Portugal version, it would be correct.

PR #812, opened 2024-09-24
Update strings.xml

azoz1492:
ID:20142736


Jeii20:
PR closed. The line you changed was actually supposed to be a period, and not to have its punctuation removed.

PR #820, opened 2024-09-26
Update strings.xml

Urtpwe:
ID: 24322008


Jeii20:
PR closed. Wrong change. Please do not insert anything on this line, because it may not work as it should.

PR #822, opened 2024-09-27
Update strings.xml

Commando70:
ID: 31784697


Jeii20:
PR closed. Too many invalid changes; you removed required punctuations, and based completely on the blue colour text at line 1334, therefore, these changes are invalid.

PR #833, opened 2024-09-28

Badyalshybany:


TurkiAlanazi1:
Cannot put space

Jeii20:
PR closed. I still advise you to read the whole video description to fulfill requests under the expected. This change is totally invalid. You should not inspire completely by the tutorial video by **providing the exact same title and description as the tutorial video**. As it says, it's an **example**, and not an order.

PR #834, opened 2024-09-28
values-PTrbr

Banned44nebu:


Jeii20:
I don't think these changes are needed, even if you're just trying to change it for "better comprehension". People can already understand witb previous texts, and your changes resulted in the same. Next time, please provide a reason for these changes and explain with enough proof why the previous texts were wrong, as it says on **Tips**:

"Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is **required**."

PR #852, opened 2024-10-03
Update strings.xml

Venomair:
My ID:29949753


zeynpcna2:
The original texts are already correct 
No change

PR #853, opened 2024-10-04
Update portuguese br laguange (example)

sansspa:
They missed a don in a string (example)


Jeii20:
PR closed. Try not keep the string length too long; previous translation was correct. You should **not** base off completely by the tutorial video, like I noticed by the title and description you provided.

PR #856, opened 2024-10-05
Indonesian Translations Improvement

arxvzy:
ID: 14517510

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -78,9 +78,9 @@
     <string name="Green_Team">"Tim Hijau"</string>
     <string name="Blue_Team">"Tim Biru"</string>
     <string name="Yellow_Team">"Tim Kuning"</string>
-    <string name="Signed_in_as_">"Di-Login sebagai:"</string>
-    <string name="Not_signed_in_">"Tidak Di-Login!"</string>
-    <string name="SET_NAME">"SET NAMA"</string>
+    <string name="Signed_in_as_">"Login sebagai:"</string>
+    <string name="Not_signed_in_">"Tidak Login!"</string>
+    <string name="SET_NAME">"ATUR NAMA"</string>
     <string name="SIGN_OUT">"LOGOUT"</string>
     <string name="ACHIEVEMENTS">"PRESTASI"</string>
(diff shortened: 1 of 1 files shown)


res/values-in/strings.xml
```diff
@@ -1261,7 +1261,7 @@ Perintah:
         <item>"Kecil"</item>
         <item>"Normal"</item>
         <item>"Besar"</item>
-        <item>"GEDE"</item>
+        <item>"Sangat Besar"</item>
```
Jeii20, line 1264: Please return the capital letters here.

Jeii20 (changes requested):

Jeii20:
Please provide a reason for these changes and explain with proof why the previous texts were incorrect, as it says on **Tips** topic:

"Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is **required**."

Jeii20:
Thanks for your time. Don't forget to provide reasons for these changes and show evidences about why they were wrong.

arxvzy:
The prefix "di-" in Indonesian is used to form passive voice for Indonesian verbs, not foreign verbs like "login." so "login" is enough instead of "di-login".

"atur nama" is common indonesian words that mean "set a name" while "set nama" is weird to read/say.

"online" in indonesian are "daring", since i saw "offline" that translated to "luring", so i edit it too for language consistency.

"upload" = "unggah", "meng-upload" is correct based in indonesian speaking, but in text to be read, "mengunggah" is better. so do the other "upload" things.

"nama warna" means like "name a color", while "warna nama" like coloring name.

"check" = "cek" in indonesia, little bit different, same sound.

"Tumpukan" is a noun, meaning "a pile" or "a stack", it can be "tumpukkan" (tumpuk-kan) or "tumpuk" is enough, same meaning, just improvement.

and lastly "GEDE", it is informal word, and "GEDE" has same meaning as "big" or "large", while "huge" is extra large and there is no indonesian word for it so we just use "sangat besar" that mean like "extra large".

all those translation based on google translate.

Jeii20:
Thanks for your time, again. ^^
Arguments and terms seem OK.

PR #864, opened 2024-10-09
Update strings.xml

27544270:


TurkiAlanazi1:
It should not be changed 
The meaning of previously is correct.

Jeiii20:
PR closed. Will result in a conflict.

PR #865, opened 2024-10-09
Update strings.xml

27544270:
ID 27544270


AnkaKoxu:
What is the difference between presence of a sign: َ
and the lack of it in the word? Is it some kind of dialect? Because you've changed the same character twice in pr, but the overall meaning doesn't change. 

PS: When you translate, please do it as a collective pull request.

27544270:
Many people have difficulty understanding its meaning, which is why I changed everything in this topic

27544270 (commented):
Translating this in order to facilitate suffering

PR #868, opened 2024-10-13
Update strings.xml

Yassin362:
Di: 27208336


TurkiAlanazi1:
I don't see any change here 🤡

Jeiii20:
Original string value should not be changed in any way (like you did in all 3 lines).

PR #875, opened 2024-10-14
Update strings.xml

27544270:
27544270


TurkiAlanazi1:
no need to change that. I don't see anything wrong. It should not be accepted
Many mistakes

27544270:
Yes mistake sorry should not be accepted

PR #876, opened 2024-10-14
Update strings.xml

zxcv1237575:
ID: 31967077


AnkaKoxu:
@TurkiAlanazi1 could you check this?

TurkiAlanazi1:
> @TurkiAlanazi1 could you check this?

I don't see any change in pr request file
Same meanings and words But it made the words bigger، should not be accepted

PR #880, opened 2024-10-15
Indonesian Update strings.xml

TakaeSlept:
Wrong Grammar and Typos

Neb ID: 17200046

Explanation:

1.Recombine = "Me-nyatukan" is way more efficient than "Bergabung Kembali" and its the same meaning

2. North South East West
         /          /         \       \
   Utara  Selatan Timur Barat

United states/US = Serikat, but we called it here as in AS (Amerika Serikat)

3. "Ter" and "Di" is not always have the same context, and the correct one is using "Ter" instead of "Di"

4. Missed Space for Login word, wrong grammar the correct  is "LOG (space) IN" same as logout

5.Dalam Jaringan/Luar Jaringan is Indonesian of Online/Offline thats it.

6. I dont know hot to explain this one but "Secara" is the correct sync for the words

7. Same meaning But way correct using "Ter" than "Di" 

8. Tampilan = shows
    Tampilkan = showing

9. "Kemenangan" Is for how many you've wins a game
    "Menang" is for you're winning one time in the game

10. Player XP= XP Pemain is the correct one

11. Double means "Ganda" in our country we can say dobel or Ganda, but its more way correct answer

12. "Ke" is for syncing the line Masuk "Ke" Arena instead Masukkan Arena means Enter An Arena

13. Remove = Hapus/Keluarkan have the same meaning depends on the context, but for this one the correct word is "Keluarkan"

I think thats it for me, its my first time making something like this, i hope you like mine, thanks anyway.

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -36,7 +36,7 @@
     <string name="_Alias_">" Nama "</string>
 
     <string name="Score__">"Skor: "</string>
-    <string name="Recombine_">"Bergabung kembali: "</string>
+    <string name="Recombine_">"Me-nyatukan: "</string>
     <string name="WINNER_">"PEMENANG:"</string>
     <string name="IDLE">"MENGANGGUR"</string>
     <string name="Name_Invalid_">"Nama tidak berlaku."</string>
@@ -53,8 +53,8 @@
     <string name="SERVER">"SERVER"</string>
     <string name="Joining_game_in_">"Bergabung permainan di "</string>
     <string name="Game_Full">"Permainan Penuh"</string>
-    <string name="US_East">"Amerika Tengah"</string>
-    <string name="US_West">"Amerika Utara"</string>
+    <string name="US_East">"AS Timur"</string>
+    <string name="US_West">"AS Barat"</string>
     <string name="Europe">"Eropa"</string>
     <string name="South_America">"Amerika Selatan"</string>
     <string name="SINGLE_PLAYER">"PEMAIN TUNGGAL"</string>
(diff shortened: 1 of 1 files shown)


AnkaKoxu:
There are a lot of substantive and good changes in this pr. 
Proposed translations texts for Offline and Online may be too long to fit into the UI. The rest are ok.

res/values-in/strings.xml
```diff
@@ -53,8 +53,8 @@
     <string name="SERVER">"SERVER"</string>
     <string name="Joining_game_in_">"Bergabung permainan di "</string>
     <string name="Game_Full">"Permainan Penuh"</string>
-    <string name="US_East">"Amerika Tengah"</string>
-    <string name="US_West">"Amerika Utara"</string>
+    <string name="US_East">"AS Timur"</string>
+    <string name="US_West">"AS Barat"</string>
```
Jeiii20, line 57: Previous texts at lines 56 and 57 were already correct (because in english is Central America and North America). Please return them.

res/values-in/strings.xml
```diff
@@ -103,9 +103,9 @@
     <string name="You">"Anda"</string>
     <string name="This_player_is_a_bot">"Pemain ini adalah bot."</string>
     <string name="This_player_is_not_signed_in">"Pemain ini tidak di-login."</string>
-    <string name="Online">"Online"</string>
+    <string name="Online">"Dalam Jaringan"</string>
```
Jeiii20, line 106: "Daring" fits much better here.

res/values-in/strings.xml
```diff
@@ -103,9 +103,9 @@
...
     <string name="Unknown">"Tidak dikenal"</string>
-    <string name="Offline">"Luring"</string>
+    <string name="Offline">"Luar Jaringan"</string>
```
Jeiii20, line 108: "Luring" was already correct, according to other native speakers. I can see they are basically an abbreviation.

Jeiii20 (changes requested):

res/values-in/strings.xml
```diff
@@ -53,7 +53,7 @@
     <string name="SERVER">"SERVER"</string>
     <string name="Joining_game_in_">"Bergabung permainan di "</string>
     <string name="Game_Full">"Permainan Penuh"</string>
-    <string name="US_East">"Amerika Tengah"</string>
+    <string name="US_East">"AS Timur"</string>
```
Jeiii20, line 56: Don't forget about this one:

```suggestion
    <string name="US_East">"Amerika Tengah"</string>

Jeiii20 (changes requested):

PR #887, opened 2024-10-25
Update strings.xml

SlaviaMadoia:
Nebulous ID 21748828


Jeiii20:
@SlaviaMadoia 

You seem to be from Hong Kong. If you would like to change your language, you can access the locations:

- `res/values-zh-rCN` (Simplified Chinese)
- `res/values-zh-rTW` (Traditional Chinese)

I am not sure which of them you speak, but they are there.

PR #890, opened 2024-10-25
German Translation Improvements

kp-leben:
id: 11462375

### DIFF
--- a/res/values-de/strings.xml
+++ b/res/values-de/strings.xml
@@ -59,8 +59,8 @@
     <string name="SERVER">"Server"</string>
     <string name="Joining_game_in_">"Spiel beitreten in "</string>
     <string name="Game_Full">"Das Spiel ist voll!"</string>
-    <string name="US_East">"USA Osten"</string>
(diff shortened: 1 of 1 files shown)


kp-leben:
I changed the cardinal points before the place... technically a direct translation of the english string, because its the correct order.

The other translation are also directly translated, as for seasonal the word could be either the word it was or saisonal because it's shorter.

To = Nach, An (An = to a person /// Nach = after / to smth)

Namensfarbe =it missed the 's' in the conjungation

Gesendete Nachrichten is correct, i changed it because of the context so it makes more sense.
Another word is Postausgang (opposite word to inbox)
((it makes more sense because you have Posteingang on the other side)

Erhalte Plasma was changed to 'Plasma erhalten' because it's more logical. (It wouldn't make a huge difference, it just looks better and is grammatically correct.

Jeiii20 (approved):
Comments are OK, thanks for your time. I will mention 2 native speakers so they can give their opinions. @0x8fv @Cedricnebu

PR #891, opened 2024-10-27
Update strings.xml

ly7788863:
ID:31969237


Jeiii20:
Please provide a reason for this change and explain with enough evidence why the previous text was wrong.

**This comment follows according to the Tips mnessage in repository's `README.md`:

- Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is **required**.

PR #892, opened 2024-10-27
Update values

Jeiii20:
## ℹ About this PR
This PR contains changes to the English language that I found interesting to make and to improve readability for English speakers at all.

## ⚙ Changes
- **Line 109:** Including a comma may help improve readability.
- **Line 110:** Same as 109.
- **Line 140:** I watched the % being removed from other languages, then decided to remove it from the English language too.
- **Line 148:** "Timed `<mode>` may improve the interpretation and users can use that structure to make changes to the languages to base off correctly. Same applies for lines 150, 308, 309, 316, 317, 338 and 342.
- **Line 382:** As of the 1st person, the "s" in the end of word was not really needed.
- **Line 355:** Same as 109.
- **Line 435:** Improved interpretation.
- **Line 465:** Inclusion of a colon would be useful in this case.
- **Line 559:** Improved interpretation.
- **Line 634:** Improved interpretation.
- **Line 685:** Improved interpretation.
- **Line 704:** Names like that may begin with the first letter as capital.
- **Line 719:** Improved interpretation.
- **Line 737:** Improved interpretation.
- **Line 739:** Improved interpretation.
- **Line 748:** More clarity for the image size limit.
- **Line 765:** "Compression Quality" doesn't make too much sense here, so in replacement to "Quality", I've used "Threshold", as we are talking about **image quality compression rate**. Might be used by users from other languages to use it to clearly match to its proposal.
- **Line 766:** Improved interpretation. Removed spaces between bars `/`, because they are not needed and may negatively affect readability.
- **Line 791:** More clarity on how to unblock someone, besides just "sending a friend request".
- **Line 839:** "YouTube" was typed at a wrong way, with the "t" in lowercase.
- **Line 899:** Improved interpretation. Considered the UI length, so split "All from" and "the clan".
- **Line 922:** Shortening sometimes is useful. May not affect what "to purchase".
- **Line 939:** Improved interpretation.
- **Line 965:** Plural required, as we do not play more than just an arena + to match with "clan war**s**".
- **Whole `Nebulous_Rules` string:** Improved interpretation. Higher clarity to encourage players on reporting someone that violates the game rules.
- **Line 988:** Improved interpretation.
- **Line 1002:** Improved interpretation.
- **Line 1012:** Improved interpretation.
- **Line 1015:** Improved interpretation.
- **Line 1031:** As Battle Royale exists, included it there.
- **Line 1037:** There is more than a place, so left "places" as plural.
- **Whole `upload_details` string:** Improved interpretation and increased clarity on the guidelines and statuses.
- **Line 1107:** The described "rule" is not appliable any longer since last year.
- **Line 1108:** Improved interpretation.
- **Line 1110:** Improved interpretation.
- **Line 1133:** Same as 766.
- **Line 1140:** Same as 109.
- **Line 1145:** Improved interpretation.
- **Line 1149:** Extra spacing removed.
- **Line 1175:** Improved interpretation.
- **Line 1176:** Improved interpretation.
- **Line 1191:** Same as 109. Improved interpretation.
- **Lines from 1200 to 1209:** As these names belong to the game modes, I made them all as capital letters.
- **Line 1214:** Same as 109.
- **Line 1215:** Same as 922.
- **Line 1225:** For those who do not know, decimal numbers (2.53, 7.49, 5.81 etc.) can also be inserted and applied, as long as they do not surpass the minimum and maximum seconds allowed.
- **Line 1227:** Improved interpretation.
- **Line 1238:** Improved interpretation.
- **Line 1261:** Same as 704. Improved interpretation.
- **Line 1274:** Same as 109.
- **Whole `skin_warning` string:** Same as 1107.
- **Line 1336:** Improved interpretation.
- **Whole `admin_commands` string:** Added 4 missing admin commands. Removed !guests.
- **Whole `sign_in_explanation` string:** Improved interpretation and higher clarity has been provided on what a signed-in user can do.
- **Whole `languages` string:** Fixed capitals from latin, cyrillic and greek alphabets.

### `<string-array name="languages">`
- **Azərbaycanlı:** This is not really correct. While "Azərbaycanlı" refers to people, "Azərbaycanca" refers to the language itself.
- **Danske:** This is not correct grammar to describe the Danish language, so changed to "Dansk". "Danske" is actually a plural from Danish.
- **Portuguese languages:** Changes from a single letter to entire name of each one of the regions for higher clarity: "Português (Brasil)" and "Português (Portugal)".

## ➕ Additional description
Change requests are greatly welcomed and appreciated. Therefore, if there is really any meaningful change to be made, I will surely make it, so please do not hesitate on it.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -106,8 +106,8 @@
     <string name="XP_">"XP:"</string>
 
     <string name="Information">"Information"</string>
-    <string name="To_use_this_skin_you_must_earn_the_achievement">"To use this skin you must earn the achievement"</string>
-    <string name="To_use_this_skin_you_must_reach_level">"To use this skin you must reach level"</string>
+    <string name="To_use_this_skin_you_must_earn_the_achievement">"To use this skin, you must earn the achievement"</string>
+    <string name="To_use_this_skin_you_must_reach_level">"To use this skin, you must reach level"</string>
     <string name="SIGN_IN">"SIGN IN"</string>
 
     <string name="Single_Player_Stats">"Single Player Stats"</string>
@@ -137,7 +137,7 @@
     <string name="Friend_Request_Sent">"Friend Request Sent…"</string>
     <string name="ERROR">"ERROR"</string>
     <string name="CONTINUE">"CONTINUE"</string>
-    <string name="Global_" formatted="false">"Global %"</string>
+    <string name="Global_" formatted="false">"Global"</string>
     <string name="FRIEND_CHAT">"FRIEND CHAT"</string>
     <string name="LOBBY_NAME">"LOBBY NAME"</string>
     <string name="NULL">"NULL"</string>
@@ -145,9 +145,9 @@
 
     <string name="ALL">"ALL"</string>
     <string name="FFA">"FFA"</string>
-    <string name="FFA_TIME">"FFA TIME"</string>
+    <string name="FFA_TIME">"TIMED FFA"</string>
     <string name="TEAMS">"TEAMS"</string>
-    <string name="TEAMS_TIME">"TEAMS TIME"</string>
+    <string name="TEAMS_TIME">"TIMED TEAMS"</string>
     <string name="CTF">"CTF"</string>
     <string name="SURVIVAL">"SURVIVAL"</string>
 
@@ -250,7 +250,7 @@
     <string name="Longest_Life_">"Longest Life:"</string>
     <string name="Games_Won_">"Games Won:"</string>
 
-    <string name="You_must_wait_until_the_end_of_the_round_">You must wait until the end of the round</string>
+    <string name="You_must_wait_until_the_end_of_the_round_">You must wait until the end of the round.</string>
     <string name="Dont_show_again">Don\'t show again</string>
     <string name="Found_a_single_player_stats_backup_with_higher_XP__Do_you_want_to_restore_the_backup_">"Found a single-player stats backup with higher XP. Do you want to restore the backup?"</string>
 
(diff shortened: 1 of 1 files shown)


AnkaKoxu:
I looked and I guess I need to book a slot here, something from me soon 😉

AnkaKoxu:
Your pr Jei is very bravely and contains quite good changes in terms of understanding some of the content, more specific and concrete information has been added, which is a plus.

## [🧠] Consider these changes
### + 834
**The pr omits the absolutely most important string to correct (834), which begins with a capital letter, which is unnecessary because it is attached to the next sentence.... “The next tournament Starts in 60 seconds.”
I would suggest a change like this:
> <string name=“starts_In”>starts in</string>.
According to this, the sentence would read: “The next tournament starts in 60 seconds".
### + 435
It seems to me that the message also pops up in the case of a purchased pack and this fact was omitted. As for the item, yes, I agree that it would be legitimate in the form of skins, pets, etc., but it is not a hit wording in the case of buying a pack (which is not an item in itself).
purchased
I would leave it as it is.
### + 786
If you put it together this way, a colon is necessary before enumerating
skins
Inappropriate skins contain**:**
### + 899
In this string, it seems to me that the dev's intention was to be as economical as possible, and I think we should stick to that, especially since the in-game space for this string is severely limited, btw, see how it looks in Polish. One line holds up to 6 characters.
clan
You have ALL FROM, which is 8 (with space)
clany
And I'm afraid Nebulous just won't withstand it xD
I would suggest leaving it as it is now...
### + 765
It was the first time I noticed the word "Threshold" in this context. And yes, I like more “Compression” than “Threshold” 
I did a lot of research.... image treshold involves converting a grayscale kub color image into a binary (black and white) image. Compression is the process of reducing the file size and the goal is to reduce the amount of space needed to store the image, wwhich fits the context better. You will find many tutorials on tresholding and yes, they all change the color, so it must be true xD So I say, I would leave it as it is, which is Image Comppresion, the threshold might not be hit-word here.
[Thresholding](https://blog.roboflow.com/image-thresholding/)
threshold


If this would be merged, I will definitely undertake the translation from 0, the pr is very promising.

Jeiii20:
Thanks for your feedback, @AnkaKoxu!!

PR #893, opened 2024-10-28
Update strings.xml

Azo13-art:
ID:20142736

PR #894, opened 2024-10-28
Update strings.xml

Azo13-art:
ID:20142736


Jeiii20:
PR closed. Same as #893

PR #895, opened 2024-10-28
Update about of strings clan name

Mompoa:
They did not find the correct tagalog of the Clan Name

### DIFF
--- a/res/values-tl/strings.xml
+++ b/res/values-tl/strings.xml
@@ -217,7 +217,7 @@
     <string name="GOAL_">"GOAL!:"</string>
     <string name="SOCCER">"SOCCER:"</string>
     <string name="DRAW">"PATAS:"</string>
-    <string name="Clan_Name">"Clan Name:"</string>
+    <string name="Clan_Name">"Pangalan ng Clan:"</string>
     <string name="Clan">"Clan:"</string>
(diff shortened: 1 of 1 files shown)


res/values-tl/strings.xml
```diff
@@ -217,7 +217,7 @@
     <string name="GOAL_">"GOAL!:"</string>
     <string name="SOCCER">"SOCCER:"</string>
     <string name="DRAW">"PATAS:"</string>
-    <string name="Clan_Name">"Clan Name:"</string>
+    <string name="Clan_Name">"Pangalan ng Clan:"</string>
```
Jeiii20, line 220: Shouldn't "clan" look like this in filipino/tagalog, according to my basic knowledge?

```suggestion
    <string name="Clan_Name">"Pangalan ng Klan:"</string>

Mompoa: That is the correct tagalog of the clan name and I am a Filipino

Jeiii20: Got it, thanks. Don't forget to mark this conversation as resolved when you can.

Jeiii20 (approved):
Translation reviewed: seems good.
Approved the changes, can merge into main as soon as possible.

PR #905, opened 2024-11-01
Update spanish languages

Oscarxxan222:
For the Spanish (Latin American) I think that the halo should be called aura, since it is something that surrounds the blob like an aura. In English it is called aura and in Spanish aura, it is the same but in English it can be left as halo and for Spanish aura

### DIFF
--- a/res/values-es/strings.xml
+++ b/res/values-es/strings.xml
@@ -976,9 +976,9 @@ Si ves a un jugador que esté rompiendo las reglas del juego, favor de denunciar
     <string name="You_cannot_open_this_present_yet_">¡No puedes abrir este regalo todavía!</string>
     <string name="Open_Present">Abrir Regalo</string>
     <string name="OPEN">ABRIR</string>
-    <string name="Halo">Halo</string>
-    <string name="To_use_this_halo_you_must_be_at_least_level_s_and_unlock_s_achievements">"Para usar este halo debes tener un nivel mínimo %1$s y desbloquear los logros %2$s."</string>
-    <string name="Halos">Halos</string>
+    <string name="Halo">Aura</string>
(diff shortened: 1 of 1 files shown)


res/values-es/strings.xml
```diff
@@ -1134,7 +1134,7 @@ Esta es una lista de cosas * PROHIBIDAS * en las skins personalizadas. Si compra
     <string name="EXCLUSIVE_skins_pets_hats_and_more_">¡Skins, Mascotas, Sombreros y Más EXCLUSIVAS!</string>
     <string name="BUY_NOW_">¡COMPRAR AHORA!</string>
     <string name="already_purchased">Tú o la persona para la que estás comprando esto ya sois dueños de esto.</string>
-    <string name="Halo_Animations">Animación del Halo</string>
+    <string name="Halo_Animations">Animación del Aura</string>
```
Jeiii20, line 1137: Just a minor grammar fix:

```suggestion
    <string name="Halo_Animations">Animación de la Aura</string>

Jeiii20 (changes requested):

Jeiii20:
Your other changes are correct.

AnkaKoxu:
[Here](https://context.reverso.net/t%C5%82umaczenie/hiszpa%C5%84ski-polski/la+aura) Aura occurs with “el”. So it's fine I think to type: del (de + el)

Jeiii20:
@AnkaKoxu 

The suffix "del" is only for masculine nouns, and "Aura" is actually feminine. ^^

AnkaKoxu:
Hmmmm
Zrzut ekranu (166)
😂

Jeiii20:
Ok, confusion has been created 💀

PR #910, opened 2024-11-04
Update Russian Language

ExenX7:
I changed the English name "PM" to the Russian name "ЛС"(Личные сообщения) And I also changed the name "Sandbox (No viruses)" to russian name "Песочница". But I didn't find this string.

29280747


Jeiii20 (approved):
About the missing string you pointed out, it still needs to be added into the file. I will add `no_virus` and all the missing strings based on `res/values` (English file) as soon as I can. Thanks for report.

PR #958, opened 2024-11-23
Update strings.xml vi from yang ho 30367133

toitenxyzzz:
Updated 1 Vietnamese word
Browse the language I translate thank.

### DIFF
--- a/res/values-vi/strings.xml
+++ b/res/values-vi/strings.xml
@@ -63,7 +63,7 @@
     <string name="South_America">"Nam Mỹ"</string>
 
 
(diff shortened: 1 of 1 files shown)


Jeiii20:
Is "Băng Đảng" really correct? According to my research, "Hội" is still correct for clan context, while "Băng Đảng" refers to a group of terrorists, as of PR #469.

toitenxyzzz:
haha because I think it would be cooler to call ''Băng đảng'' cool than ''HộI'' because Hội is just a normal phrase that most of us Vietnamese people call ''BĂNG ĐẢNG''or ''Gia tộc''

PR #981, opened 2024-12-16

Ahmed1540:

### DIFF
--- a/res/values-ar/strings.xml
+++ b/res/values-ar/strings.xml
@@ -63,7 +63,7 @@
     <string name="South_America">"امريكا الجنوبية"</string>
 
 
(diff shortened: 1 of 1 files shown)


Jeiii20:
Please provide a reason for this change and explain with enough evidence why the previous text was wrong.

**This comment follows as it's shown up in Tips topic, at [main repository page}(https://github.com/simplicialsoftware/nebulous-translations):**

- "Adding comments to your PR that explains why subtle changes were made can help. If you are revising something that is already translated, an explanation is **required**."

-----
You should not base off completely by the tutorial video; I am talking about the title and description you provided for this PR. As it says, it's an **example**, and not an order.

PR #991, opened 2024-12-27
Update Indonesia Language

al-qassab:
they not translate it yet in string 1096

id: 22963766
nickname: zexun
level: 1168


al-qassab:
id: 22963766

Jeiii20 (approved):
Translation reviewed and it seems accurate.

PR #993, opened 2024-12-29
Update indonesia asia language (example)

saruldong:
They are missing a colon in a string

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -95,7 +95,7 @@
     <string name="FRIENDS">"TEMAN"</string>
     <string name="Warning">"Peringatan"</string>
     <string name="SIGNING_IN">"LOGIN…"</string>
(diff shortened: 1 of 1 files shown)


res/values-in/strings.xml
```diff
@@ -95,7 +95,7 @@
     <string name="FRIENDS">"TEMAN"</string>
     <string name="Warning">"Peringatan"</string>
     <string name="SIGNING_IN">"LOGIN…"</string>
-    <string name="New_Friend_Request">"Permintaan Teman Baru"</string>
+    <string name="New_Friend_Request">"Permintaan Teman Baru:"</string>
```
Jeiii20, line 98: It should be an exclamation mark here.

```suggestion
    <string name="New_Friend_Request">"Permintaan Teman Baru!"</string>

Jeiii20 (changes requested):

PR #995, opened 2024-12-29
Indonesian language update

Khoirun0524:
They forgot to remove the dot on the string

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -11,17 +11,17 @@
     <string name="WEEKLY">"MINGGUAN"</string>
     <string name="DAILY">"HARIAN"</string>
     <string name="Server_">"Server:"</string>
-    <string name="Game_Mode_">"Mode Permainan:"</string>
+    <string name="Game_Mode_">"Mode Permainan"</string>
     <string name="Show_Names">"Tampilkan Nama"</string>
     <string name="Show_Avatars">"Tampilkan Skin"</string>
     <string name="PLAY">"MAIN"</string>
     <string name="OPTIONS">"OPSI"</string>
 
-    <string name="Avatar_">"Skins:"</string>
-    <string name="Name_">"Nama:"</string>
+    <string name="Avatar_">"Skins"</string>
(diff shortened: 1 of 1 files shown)


res/values-in/strings.xml
```diff
@@ -11,17 +11,17 @@
     <string name="WEEKLY">"MINGGUAN"</string>
     <string name="DAILY">"HARIAN"</string>
     <string name="Server_">"Server:"</string>
-    <string name="Game_Mode_">"Mode Permainan:"</string>
+    <string name="Game_Mode_">"Mode Permainan"</string>
```
Jeiii20, line 14: Please return the colon here.

res/values-in/strings.xml
```diff
@@ -11,17 +11,17 @@
...
     <string name="Show_Names">"Tampilkan Nama"</string>
     <string name="Show_Avatars">"Tampilkan Skin"</string>
     <string name="PLAY">"MAIN"</string>
     <string name="OPTIONS">"OPSI"</string>
 
-    <string name="Avatar_">"Skins:"</string>
-    <string name="Name_">"Nama:"</string>
+    <string name="Avatar_">"Skins"</string>
+    <string name="Name_">"Nama"</string>
```
Jeiii20, line 21: Please return the colons here.

Jeiii20 (changes requested):

res/values-in/strings.xml
Khoirun0524: this needs to be translated
Screenshot_2024-12-30-11-53-14-718_software simplicial nebulous

Khoirun0524: item PLASMA_BOOST_3X_1D_XMAS/0 not found.

item PLASMA_BOOST_3X_1D_NATAL/0 tidak ditemukan.

Jeiii20: @Khoirun0524

Item names are not supposed to be translated. If you have some untranslated text to report, please use the [issue tracker](https://github.com/simplicialsoftware/nebulous-translations/issues).

PR #997, opened 2024-12-30
Update indonesia asia language

Satoru01nihbro:
Adding an exclamation mark

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -93,9 +93,9 @@
     <string name="Level">"Level"</string>
     <string name="PAUSED">"DIJEDA"</string>
     <string name="FRIENDS">"TEMAN"</string>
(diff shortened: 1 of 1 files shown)


Satoru01nihbro:
id : 31724132

res/values-in/strings.xml
```diff
@@ -93,9 +93,9 @@
     <string name="Level">"Level"</string>
     <string name="PAUSED">"DIJEDA"</string>
     <string name="FRIENDS">"TEMAN"</string>
-    <string name="Warning">"Peringatan"</string>
+    <string name="Warning">"Peringatan!"</string>
```
Jeiii20, line 96: No any exclamation needed here. Keep in mind that translations must be based at English file (`res/values`).

Jeiii20 (changes requested):

PR #998, opened 2024-12-30
Indonesian translation update

Khoirun0524:


res/values-in/strings.xml
```diff
@@ -1160,7 +1160,7 @@
     <string name="BATTLE_ROYALE">PERTEMPURAN ROYALE</string>
     <string name="Battle_Royale">Pertempuran Royale</string>
     <string name="The_winner_will_receive_x_plasma_">Pemenang akan mendapatkan %s plasma!</string>
-    <string name="admin_commands">"
+    <string name="admin_commands">"perintah admin"</string>
```
Jeiii20, line 1163: Please remove this text. `</string> already appears at line 1180.

```suggestion
    <string name="admin_commands">

Jeiii20 (changes requested):

PR #1006, opened 2025-01-04
Portuguse BR translation improvements

SanikYT:
ID:30100094


Jeiii20:
@SanikYT 
PR closed because it contains wrong change.
Portuguese-speaker users can already understand by "Máx. de Pessoas" and is the correct phrase.
You seem to be Arabic-native, then for that you must change `res/values-ar`.

PR #1026, opened 2025-01-16
update indonesia in language

rul1824:
English to Indonesian translation update

my id : 31917581


res/values-in/strings.xml
```diff
@@ -846,9 +846,9 @@
     <string name="error_1">Hanya Pemimpin Atau Wakil Pemimpin Yang Bisa Mengirim Mail Klan.</string>
     <string name="Hat">Topi</string>
     <string name="Hats">Topi</string>
-    <string name="Pet_Uploads">Pet Upload</string>
+    <string name="Pet_Uploads">Unggahan Hewan Peliharaan</string>
     <string name="BLOB">GUMPALAN</string>
-    <string name="Clan_Pet_Uploads">Klan Skin Pet Upload</string>
+    <string name="Clan_Pet_Uploads">Unggahan Hewan Peliharaan Klan</string>
```
TheLilUser, line 851: Would be good if you try to abbreviate this string because it's too long compared to English.

PR #1030, opened 2025-01-18
Update Español (example)

JAUSSJSNDUJ:
ID: 32394642 NEBULOUS.IO


TheLilUser:
The changes just contain removals of quotation, which was not needed.

PR #1033, opened 2025-01-18
Update Portuguese Br language

Rayanny07:


res/values-pt-rBR/strings.xml
```diff
@@ -60,10 +60,10 @@
     <string name="US_East">"América Central"</string>
     <string name="US_West">"América do Norte"</string>
     <string name="Europe">"Europa"</string>
-    <string name="South_America">"Brasil"</string>
+    <string name="South_America">"America do Sul"</string>
```
TheLilUser, line 63: Two issues:

1. Missing diacritic in the word "América".
2. Though "América do Sul" is the correct translation to "South America", Brazilian users may be confused if the server name changes, as they got used to the current name.

```suggestion
    <string name="South_America">"Brasil"</string>

TheLilUser (changes requested):

TheLilUser:
About the "ESPERE", it doesn't really seem right for the context and may negatively affect the formality.

PR #1035, opened 2025-01-19
Aabic AR translations Impravements

iyads7:
ID:32087443


TheLilUser:
According to a native, your changes are wrong.

PR #1038, opened 2025-01-22
Update strings.xml

Sayussss:
Make it easier to understand.


Sayussss:
I need tag translator ID 23233003

PR #1053, opened 2025-01-30

ysh06379:


TheLilUser:
Hi @ysh06379 
You should not base off completely by the tutorial video. Please make sure you are changing `res/values-ar` file, and not `res/values-pt-rBR`.

TheLilUser:
Inaccurate change.

PR #1055, opened 2025-02-03
Update translations in Spanish (ES)

Santiago926:
A spelling correction was made on line number 312. One of the words was missing a letter.

the correction was " in " = " Sin "

A correction was also made to line 533, the text was missing initial quotes.

And finally, a correction was also made on line 706, it was also missing initial quotes from the text

ID: 19040794


Santiago926:
A suggestion that I could add is to correct grammatically and so that it sounds more formal is the following:

<string name="Stick_Size_">"Tamaño pad:"</string>
    <string name="Stick_Margin_">"Ubicación pad:"</string>

Add " del " in these 2 lines of code so that it is understood as follows:

<string name="Stick_Size_">"Tamaño del pad:"</string>
    <string name="Stick_Margin_">"Ubicación del pad:"</string>

Santiago926:
I will continue doing more review of the String

PR #1063, opened 2025-02-09
Portuguese BR Translate Improvements

Ryzen7-source:
Nebulous id: 14776768

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -381,7 +381,7 @@
     <string name="Spectating">"Espectador"</string>
 
 
-    <string name="snowflakes_">"flocos de neve."</string>
+    <string name="snowflakes_">"Flocos de neve."</string>
     <string name="presents_">"presentes."</string>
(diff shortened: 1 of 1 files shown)


res/values-pt-rBR/strings.xml
```diff
@@ -602,7 +602,7 @@
 
     <string name="Facebook_Profile">"Perfil do Facebook"</string>
     <string name="You_must_be_signed_in_to_Facebook_">"Você precisa estar conectado ao Facebook."</string>
-    <string name="copy_details">"Copiar Detalhes"</string>
+    <string name="copy_details">"copiar detalhes"</string>
```
TheLilUser, line 605: It's not because it's entirely small letter in original string value that it should be small letter in Portuguese (or whatever language) as well.

```suggestion
    <string name="copy_details">"Copiar Detalhes"</string>

TheLilUser (changes requested):

TheLilUser (approved):
Thanks for your contribution.
Approved the changes, can merge into main as soon as possible.

PR #1070, opened 2025-02-09
Update strings.xml

Yujiro2:
Sounds better in German and is shorter 

ID: 20220113


TheLilUser:
According to Cedric, this change is inaccurate.

PR #1079, opened 2025-02-14
Update Arabic br Ianguage (example)

Hdudb134:
The missed a dot in a string (example)


AnkaKoxu:
Is it true that your proposed word is mainly used for sexual harassment? If so, it's better not to change it because reports with harassment applies to all types of harassment, not just sexual harassment.

TheLilUser:
Change reviewed by a native, should not be merged.

PR #1080, opened 2025-02-14
Update Arabic br Ianguage (example)

Hdudb134:
They missed a dot in a string
(example)
Id : 32014304


AnkaKoxu:
I'm not sure it is neccesary to change SKIN meaning in one of many strings.

TheLilUser:
Change reviewed by a native, should not be merged.

PR #1081, opened 2025-02-14
Update Lenguaje español

Sunsetz3:
ID : 22999378


TheLilUser:
Changes reviewed and too many of them are inaccurate + original string value (in blue colour) has been changed in too many lines, should not be merged.

PR #1099, opened 2025-02-27

nebullosq:


TheLilUser:
Contains changes to original string value, should not be approved nor merged.

PR #1100, opened 2025-02-28
Update Español - Castellano

Daniel-007-sm:
La mayoría de las personas latinas le decimos al control = "Mandó"


TheLilUser:
Change reviewed by CritiK and doesn't seem accurate, should not be approved nor merged.

PR #1105, opened 2025-02-28
Update español - Lenguaje castellano

Daniel-007-sm:
Muchos no saben que significa "skins"


TheLilUser:
Inaccurate change.

Please do not translate it too literally. "Skin" is already globally used for in-game skin, and your description is not really right.

PR #1108, opened 2025-02-28
Update español - Lenguaje castellano

Daniel-007-sm:
Una palabra con mucho más sentido


TheLilUser:
"Bot" and "Robot" have the exactly same meaning, should keep the previous word in my opinion.

PR #1133, opened 2025-03-25
  Update Ukrainian uk language

marnota1:
(TEST)

I tried to make an even better translation. 
My ID: 6297943
My Discord account: marnota


AnkaKoxu:
Tycoon achievement should be translated using capital letters. Search for the english, original version.

marnota1:
> Tycoon achievement should be translated using capital letters. Search for the english, original version.

Thank you for your excuses. You're right.

PR #1139, opened 2025-04-05

yusef357:


yusef357:
You should add an explanation video like the one I saw in Arabic. It was very easy now, but because I did not understand what I had to do, I did not correct all the mistakes. 🤧

John-Oz:
Changed original values, cannot be approved nor merged.

PR #1167, opened 2025-04-21
Update strings.xml

27544270:
ID: 27544270


AnkaKoxu:
Hi. Why so change?

27544270:
In order to clarify the meaning of the word and make it easier to understand.

PR #1170, opened 2025-04-23
Update Ukrainian uk language

marnota1:
I already have a translator role and want to update my translation. Discord: @marnota 
ID: 6297943


AnkaKoxu:
It might be too long to match with the UI but I'm not sure.

PR #1171, opened 2025-04-24
Update Nome Vietnamese

Wxrror:
Missed a dot


AnkaKoxu:
This change is invalid.

PR #1174, opened 2025-04-26
Azerbaijani az language typo correction

Shah-Samir:
Button texts should always be in command form and should be written in capital letters  ID 28362880


AnkaKoxu:
Good changes.

PR #1175, opened 2025-04-28
Update strings.xml (PT/BR)

Alysson-Js:


AnkaKoxu:
Under no circumstances should you change texts on the left, you can do only ones in the right.

PR #1177, opened 2025-04-28
Azerbaijani AZ Translation Improvements

JackSolomon622:
my ID: 22407446


JackSolomon622:
> Hi. Could you explain this change?

hi. i just changed some words that was false

AnkaKoxu:
You're right👍🏻

PR #1184, opened 2025-04-29
Update English usa Linguagem (exemplo)

JhulianNeb:
They missed a Dot in a string (example)


AnkaKoxu:
As far as I can see, Spanish thongs do not use dots after the names of similiar things, this change is invalid.

Zrzut ekranu (4)
Zrzut ekranu (3)

etc...

PR #1188, opened 2025-04-30
Update lenguaje español

3022741719:
Corregí 3 puntos 

ID 23117700


res/values-b+es+r419/strings.xml
```diff
@@ -942,7 +942,7 @@
     <string name="SET_NOTE">ESTABLECER NOTA</string>
     <string name="enable_notifications">habilitar notificaciones</string>
     <string name="PROMOTE_MEMBER">ASCENDER MIEMBRO</string>
-    <string name="DEMOTE_MEMBER">DEGRADAR MIEMBRO</string>
+    <string name="DESCEND_MEMBER">DESCENDER MIEMBRO</string>
```
AnkaKoxu, line 945: You can not change texts on the left. This string is ok, second thing is you shouldn't change it in any way. Demote member as degradar miembro is perfect translation for it.

PR #1190, opened 2025-04-30
Español ES Translation improvements

HawkE25:
19436414


AnkaKoxu:
It's hard to say if this is a good change, it would be possible to determine if there were more changes, because from what I've glanced, spanish strings in the settings use lowercase letters rather constantly.
Bez nazwy

PR #1191, opened 2025-04-30
Update portuguese br language

JhulianNeb:
My id: 32739880


AnkaKoxu:
Invalid change - space is not required here

PR #1194, opened 2025-04-30
Update English usa Linguagem (exemplo)

JhulianNeb:
They missed a dot in a string in My id 32739880


AnkaKoxu:
@JhulianNeb please, stop insist on making invalid changes. They are not ok. consider reading translation tips & instructions instead.

PR #1198, opened 2025-05-04
Czech Translation Improvements

Plastic-svg:
17952898 thanks, dunno exactly the context, sorry if its wrong, but in czechian it is like that


AnkaKoxu:
Hi. If you're asking about the context of this word, which is crucial, you can glance this on the leaderboard ingame (daily, weekly, all).

It means **weekly leaderboard**.

PR #1199, opened 2025-05-04
Update russian language

Boarzz771:
HUGE map size was changed on Огромное

22726241

### DIFF
--- a/res/values-ru/strings.xml
+++ b/res/values-ru/strings.xml
@@ -1389,7 +1389,7 @@
         <item>"Маленькое"</item>
         <item>"Нормальное"</item>
         <item>"Большое"</item>
-        <item>"HUGE"</item>
+        <item>"Огромное"</item>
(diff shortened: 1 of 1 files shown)


Majaaaa1:
This is ok, it can be merged into main.

Majaaaa1:
I wanted to write that, but look, there's nothing in the file for tinny, small, etc maybe you could change it

res/values-ru/strings.xml
```diff
@@ -1389,7 +1389,7 @@
         <item>"Маленькое"</item>
         <item>"Нормальное"</item>
         <item>"Большое"</item>
-        <item>"HUGE"</item>
+        <item>"Огромное"</item>
```
AnkaKoxu, line 1392: This should be written using capital letters.

AnkaKoxu (changes requested):

PR #1201, opened 2025-05-05
Modernizing the Arabic language (example)

zain11129:
You missed a number of other points and things I pointed out, feel free to check them out.


zain11129:
ID (32712826)

AnkaKoxu:
Unfortunately most of these changes are incorrect, especially when it comes to adding dots.

PR #1202, opened 2025-05-05
Update Spanish br language (example)

sergio13p:
I want to be a translator and I know English and Spanish. (example)

### DIFF
--- a/res/values-pt-rPT/strings.xml
+++ b/res/values-pt-rPT/strings.xml
@@ -1,4 +1,4 @@
-<resources xmlns:tools="http://schemas.android.com/tools">
+<recursos xmlns:tools="http://schemas.android.com/tools">
     <string name="Players">"Jogadores"</string>
     <string name="Control_">"Controlo:"</string>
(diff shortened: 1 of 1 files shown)


sergio13p:
It's okay, I hope I can be a translator. I'll do it again. If it's not correct, it's because I didn't understand the tutorial very well.

sergio13p:
Id 20359139

AnkaKoxu:
Read it again.

sergio13p:
I read it again and see what is wrong and correct it?

AnkaKoxu:
@sergio13p I glanced at it and you shouldn't changes text on the left, you can do it only with those right ones.

Majaaaa1:
@sergio13p Please pay attention to the important rules. If you are not clear with the video we sent and the guide here then I'm afraid you won't be able to do this correctly. @AnkaKoxu has already pointed out the error to you, there is no need to open a ticket until your PR is good according to the rules and merged into main.

PR #1213, opened 2025-05-08
Arabic translation

HusseinPro80:
Id 24972103


Akumakg (approved):

TurkiAlanazi1:
The file consists of a collection of dots between words, and I don't think it should be accepted.

mody515 (commented):
He did not make any fundamental changes, only formal things. The periods and punctuation marks should not be deleted. This is wrong.

PR #1218, opened 2025-05-16
Update translations in Spanish V2 (ES) .xml

Santiago926:
Changes were made in spelling, grammar, and reasoning. Since I made many changes, I will explain each correction and explain why I made those changes. I will also suggest grammar changes in case they let me apply them in the next commit.

Changes were made to quotation marks and also to the use of upper and lower case letters.

1. Change one

In the change from line one to line 5, the words "de" and "del" are added to make the sentence from English to Spanish sound more complete. 

2.Change two

On line 38, we make a more beautiful translation from English to Spanish, "Recombine" Although the Spanish translation is "Juntar" (To get together), the Spanish translation "Recombinar" (Recombining) is more natural.

3.Change three

Line 154 and 155 The way the achievement is said has been changed to "Devorador de puntos" to "Devorapuntos" Which sounds a little more fun and shortened to say the achievement and looks distinctive to the other achievements and also changed "Goloso" to "Glotón"  Which best resembles "Gormandizer"

4.Change four 

Change was made in the line 171 since the word "Blobocalipsis" translated into Spanish as Apocalypse does not have an accent on the i

5. Change five

In this change, only a space is added at the end of the sentence, respecting the name tag at the end.

6. Change six

In line 249 the word "Blob" is left with the initial capital letter because it is defined in English as a proper noun, and in Spanish blob is also used as a proper noun, since it is in lowercase.

7. Change seven

On line 289 the phrase is changed from "Dicho" to "ningún" since it is referring to the fact that no one was found  user, since the word "" sounds very open and the translation sounds less natural 

Generate a Version 3 (V3) with the exact corrections and more language corrections, the changes I made do not change the way of expression from English to Spanish, it is about maintaining the translation

ID NEBULOUS: 19040794

Changes for the Traslate tag

### DIFF
--- a/res/values-es/strings.xml
+++ b/res/values-es/strings.xml
@@ -2,10 +2,10 @@
     <string name="Players">"Jugadores"</string>
     <string name="Control_">"Control:"</string>
     <string name="CLOSE">"CERRAR"</string>
-    <string name="Stick_Size_">"Tamaño pad:"</string>
-    <string name="Stick_Margin_">"Ubicación pad:"</string>
-    <string name="Button_Size_">"Tamaño botones:"</string>
-    <string name="Button_Margin_">"Margen botones:"</string>
+    <string name="Stick_Size_">"Tamaño del pad:"</string>
+    <string name="Stick_Margin_">"Ubicación del pad:"</string>
+    <string name="Button_Size_">"Tamaño de botones:"</string>
+    <string name="Button_Margin_">"Margen de botones:"</string>
     <string name="Buttons_">"Botones:"</string>
     <string name="DEFAULTS">"POR DEFECTO"</string>
     <string name="WEEKLY">"SEMANAL"</string>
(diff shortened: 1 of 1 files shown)


res/values-es/strings.xml
```diff
@@ -18,7 +18,7 @@
     <string name="OPTIONS">"OPCIONES"</string>
 
     <string name="Avatar_">"Skin:"</string>
-    <string name="Name_">"Nombre:"</string>
+    <string name="Name_">"Nombre: "</string>
```
Santiago926, line 21: Ignore this change

res/values-es/strings.xml
```diff
@@ -64,8 +64,8 @@
 
 
     <string name="SINGLE_PLAYER">"UN JUGADOR"</string>
-    <string name="Background_">"Fondo:"</string>
-    <string name="Difficulty_">"Dificultad:"</string>
+    <string name="Background_">"Fondo: "</string>
+    <string name="Difficulty_">"Dificultad: "</string>
```
Santiago926, line 68: Ignore this change

PR #1223, opened 2025-05-17
Arabic translation

bmn521:
id- 2940898
line 796 pandemic = جائحة


mody515 (approved):
Good, pandemic= جائحة not 
زومبي=Zombie 
I think it should be approved.

PR #1226, opened 2025-05-17
Change some of the words of the German language

ahmdhjknj:
Hello, I have modified some German words to be more common for the player who reads the first change is the word In line 237 I changed the word (Masse) and replaced it with (Menge) because I found it more common.  2 In line 133 I changed the word Global and replaced it with Weltweit because I found it more popular and aesthetically pleasing. In line 131 I changed the word Fehler to Fehler because I found it more common and because it is a word that is frequently used. Problem because I find it more common and because it is a word that is used a lot In line 113 I changed the word Warnung to Hinweis because I find it beautiful and many people use it  In line 112 I changed the word Freunde to Freundesliste because it expresses the place where friends are found and they can be unfriends because I add them to take a temporary favor from them and they are considered unfriends.

### DIFF
--- a/res/values-de/strings.xml
+++ b/res/values-de/strings.xml
@@ -67,7 +67,7 @@
 
     <string name="SINGLE_PLAYER">"Einzelspieler:"</string>
     <string name="Background_">"Hintergrund:"</string>
-    <string name="Difficulty_">"Schwierigkeit:"</string>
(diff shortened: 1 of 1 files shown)


ahmdhjknj:
ID 32712826

mody515 (approved):

PR #1231, opened 2025-05-19
[5032837] Fix german translations

0x8fv:
Reverted and improved some of the translations.
Voted with Maja in the German text channel on Discord.


AnkaKoxu:
Second person form fits better. You're right here.

PR #1237, opened 2025-05-23
Indonesia ( IN ) Improvements Translation

7Shann:
My ID : 14001294

The reason I made the translation improvements was so that the language used in this game would be easier to understand for people of all ages, Many of the languages used previously were a little less easy to understand, so since the majority of nebulous game players are mostly under 20 years old, Or you could say that many of the users are teenagers and adults, so I took the initiative to create a language that is easy enough to recognize and understand for all age groups. That's all and thank you!

### DIFF
--- a/res/values-in/strings.xml
+++ b/res/values-in/strings.xml
@@ -24,7 +24,7 @@
     <string name="Connecting___">"Menghubungkan …"</string>
 
     <string name="Are_You_Sure_">"Apakah Anda yakin?"</string>
-    <string name="Yes">"Ya"</string>
+    <string name="Yes">"Iya"</string>
     <string name="No">"Tidak"</string>
     <string name="WAIT">"TUNGGU…"</string>
     <string name="Must_Specify_Name_">"Harus menentukan nama."</string>
@@ -60,7 +60,7 @@
     <string name="SINGLE_PLAYER">"PEMAIN TUNGGAL"</string>
     <string name="Background_">"Latar Belakang:"</string>
     <string name="Difficulty_">"Kesulitan:"</string>
-    <string name="Delay_Disconnect">"Memperlambat Disconnect"</string>
+    <string name="Delay_Disconnect">"Memperlambat Terputus"</string>
     <string name="Max_Players">"Maks Pemain"</string>
     <string name="Min_Players">"Min Pemain"</string>
     <string name="OK">"OKE"</string>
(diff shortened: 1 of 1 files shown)


res/values-in/strings.xml
```diff
@@ -124,20 +124,20 @@
...
-    <string name="Dot_Gormandizer">"Dot Gormandizer"</string>
-    <string name="Dot_Sommelier">"Dot Sommelier"</string>
+    <string name="Dot_Nibbler">"Titik Nibbler"</string>
+    <string name="Dot_Muncher">"Titik Muncher"</string>
+    <string name="Dot_Eater">"Titik Eater"</string>
+    <string name="Dot_Gulper">"Titik Gulper"</string>
+    <string name="Dot_Gobbler">"Titik Gobbler"</string>
+    <string name="Dot_Devourer">"Titik Devourer"</string>
+    <string name="Dot_Gormandizer">"Titik Gormandizer"</string>
+    <string name="Dot_Sommelier">"Titik Sommelier"</string>
```
AnkaKoxu, line 134: I have a loose thought, wouldn't it be better to translate the achievements into Indonesian without leaving these Nibbler, Muncher, Gulper etc. in english?

res/values-in/strings.xml
```diff
@@ -124,20 +124,20 @@
...
     <string name="Quadra_Gulp">"Quadra Gulp"</string>
     <string name="Multi_Gulp">"Multi Gulp"</string>
     <string name="Mega_Gulp">"Mega Gulp"</string>
     <string name="Ultra_Gulp">"Ultra Gulp"</string>
     <string name="Monster_Gulp">"Raksasa Gulp"</string>
-    <string name="Ludicrous_Gulp">"Menggelikan Gulp"</string>
+    <string name="Ludicrous_Gulp">"Gulp Menggelikan"</string>
```
AnkaKoxu, line 140: same here, wouldn't you translate Gulp? (if it's posibble)

res/values-in/strings.xml
```diff
@@ -1193,7 +1193,7 @@ Perintah:
     <string name="Complete_campaign_mission_">Selesaikan kampanye misi</string>
     <string name="Specify_Reason_">Alasan spesifik:</string>
     <string name="FREE_FOR_ALL">GRATIS UNTUK SEMUA</string>
-    <string name="SPECIAL">SPESIAL</string>
+    <string name="SPECIAL">Istimewa</string>
```
AnkaKoxu, line 1196: Capital letters are required here.

PR #1244, opened 2025-05-25
Update strings.xml

andyfeez7777:
ID: 23164256


manu12223:
Invalid translation: User changed "membri" to "member". You should translate the english strings to italian, not the italian back to english

PR #1247, opened 2025-05-25
Update strings.xml

andyfeez7777:
ID: 23164256


AnkaKoxu:
Under no circumstances should you change texts on the left. You can only edit texts on the right.

PR #1225, opened 2025-05-17
Update strings.xml

manu12223:
ID: 1196782

PR #1301, opened 2025-06-12
Update strings.xml

MINHAZ1292:
ID:32283639


manu12223:
Invalid translation
"vinciriceverà" is not an Italian word and this line was already suggested to be fixed in https://github.com/simplicialsoftware/nebulous-translations/pull/1225.

PR #1311, opened 2025-06-14
Update strings.xml

Thegoodboy299:
ID:13169356


Majaaaa1:
@Thegoodboy299 You need to be patient until your request is merged into main, only then open a ticket. Of course, a tag requires at least 10 changes.

PR #1337, opened 2025-06-27
Update strings.xml

KeinMussolini:
ID: 33013900


manu12223:
#1225 copied tutorial video

PR #1357, opened 2025-07-04
Update strings Bahasa Indonesia.xml

Ajaxs69:
Request tags translator
Nebulous lo account name: AJAXS
ID: 26500324
I have translated the wrong language into correct Indonesian.


Ajaxs69:
Thank you for the help

PR #1381, opened 2025-07-14
Update strings.xml

hazard-178:
ID 31343544


MakotoJohn:
Can close.

**In Italian:**

- "Finale" (ends with "-e") is singular.
- "Finali" (ends with "-i") is plural.

PR #1392, opened 2025-07-14
Update strings.xml

hazard-178:
"Ok" es de acuerdo
ID:31343544


MakotoJohn:
Can close; unnecessary change to an excessively long text. Plus "VALE" is widely known and better since it's short, readable and optimiser for UI.

PR #1395, opened 2025-07-14
Update strings.xml

hazard-178:
Estaba mal escrito, es reporte o informe.
ID:31343544

### DIFF
--- a/res/values-es/strings.xml
+++ b/res/values-es/strings.xml
@@ -552,7 +552,7 @@
 
     <string name="Invited">"Invitado"</string>
     <string name="Invitation_Sent">"Invitación enviada"</string>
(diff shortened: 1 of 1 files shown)


res/values-es/strings.xml
```diff
@@ -552,7 +552,7 @@
 
     <string name="Invited">"Invitado"</string>
     <string name="Invitation_Sent">"Invitación enviada"</string>
-    <string name="REPORT">"DENUNCIAR"</string>
+    <string name="REPORT">"REPORTE"</string>
```
MakotoJohn, line 555: I think this would be better, since it is a text that appears in a button:

```suggestion
    <string name="REPORT">"REPORTAR"</string>

PR #1415, opened 2025-07-23
Update strings.xml Indonesian Language Translation Improvement 

Gabrielaloveayato:
My ID : 5533826 (gab~licious) asia server

Thank you mods🌷


MakotoJohn:
Can close; changed texts at the left (not allowed without a valid reason).

PR #1428, opened 2025-07-27
Update Tr Translation

Dom-dim:
ID: 3333789


zeynpcna2 (commented):
Correct translate can be merge

Dom-dim (commented):
ID: 3333789

PR #1433, opened 2025-07-29
Update Tr Translation

Dom-dim:
ID: 3333789


Mochbadar06:
What do u mean?

Dom-dim:
> Ne demek istediniz?

Did I say something to you?

PR #1443, opened 2025-08-02
Update strings.xml

shirbits:
I have modified from English to Simplified Chinese.  My ID: 31850631


manu12223 (changes requested):
The property 'translatable' must be either true or false. Do not translate the words true or false.

MakotoJohn:
Can close.

PR #1452, opened 2025-08-05
Update strings.xml ES

agustin638847gg:
ID: 29521850


AnkaKoxu:
These changes are invalid, bcs:
1) you've removed placeholder: %s, which is crucial to display user properly. And placeholder there is not a mistake.
2) you've removed a space in string ALL_CLAN, which it might cause the text block to be displayed incorrectly in the UI.

MakotoJohn:
Can close.

PR #1455, opened 2025-08-07
Update strings.xml Translations

soumenhalder5228-gmail:
ID 32258047


AnkaKoxu:
Nuh uh, this change is invalid. As far I can see, you're trying to make two identical lines of translation. Do not change the lines on the left side, only on the right.
What do you want to do exactly?

MakotoJohn:
Can close.

PR #1490, opened 2025-09-17
Update strings.xml

unknown549823-svg:
ID-29336353 i did change in 689 kuu kugoja means moon collector


AnkaKoxu:
Under no circumstanses should you change texts on the left. You can only edit right ones.

AnkaKoxu:
Under no circumstances should you change texts on the left. You can only edit right ones.

John-Ozz:
At this point you should just change the translation from "Kuu Koguja" to "Kuu Kollektsionäär".

MakotoJohn:
Can close.

PR #1503, opened 2025-09-24
Update strings.xml

Rom-hub-77:
ID:13039444


AnkaKoxu:
Nieprawidłowy ID Konta ✅️
Nieprawidłowe ID Konta ❌️

Example of Google text with same meaning:
Screenshot_20250924_191501_Chrome
In my opinion, this change is not correct, although ID is an abbrevation (for UI purposes), so the grammar form should matched with full word (identyfikator) 🇵🇱

Comp banned is ok, but I think this might be too long (bcs of the UI).

PR #1513, opened 2025-10-04
Update strings.xml

Rom-hub-77:
ID:13039444

### DIFF
--- a/res/values-pl/strings.xml
+++ b/res/values-pl/strings.xml
@@ -297,7 +297,7 @@
     <string name="Wins">"Wygrywają"</string>
     <string name="Score">"Wynik"</string>
     <string name="Rating">"Ocena"</string>
-    <string name="You_are_about_to_make_a_new_leader_">"Czy na pewno chcesz mianować tą osobę Liderem Klanu?\nUważaj! Ta opcja jest bardzo ryzykowna."</string>
+    <string name="You_are_about_to_make_a_new_leader_">"Czy na pewno chcesz mianować tę osobę Liderem Klanu?\nUważaj! Ta opcja jest bardzo ryzykowna."</string>
     <string name="XP">"XP"</string>
     <string name="High_Score">"Wysoki wynik"</string>
     <string name="_3v3">"3vs3"</string>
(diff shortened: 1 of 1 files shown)


res/values-pl/strings.xml
```diff
@@ -612,7 +612,7 @@
     <string name="contribution_warning">"Czy na pewno chcesz wpłacić tę plazmę? Nie ma możliwości jej zwrotu."</string>
     <string name="Upload">"PRZEŚLIJ"</string>
     <string name="upload_warning">"Czy na pewno chcesz przesłać tę skórkę tak, aby inni gracze mogli ją zobaczyć? Rozpatrzenie skórki może potrwać do tygodnia. Nieodpowiednie skórki będą odrzucane i nie będą zwracane!"</string>
-    <string name="In_Review">"W Toku Rozpatrzenia"</string>
+    <string name="In_Review">"W trakcie rozpatrywania"</string>
```
AnkaKoxu, line 615: First letters need to be capitalised as in the original values:
W Trakcie Rozpatrywania

res/values-pl/strings.xml
```diff
@@ -388,7 +388,7 @@
     <string name="Win_100_Arenas_">"Wygraj 100 Aren."</string>
     <string name="Win_1000_Arenas">"Wygraj 1.000 Aren."</string>
     <string name="Already_in_arena_">"Jesteś już na Arenie."</string>
-    <string name="Already_in_queue_">"Już jesteś w kolejce."</string>
+    <string name="Already_in_queue_">"Jesteś w kolejce."</string>
```
AnkaKoxu, line 391: You've translated this as:
Jesteś w kolejce. = You are in queue - and I find it wrong.
It should be noted that the player is ALREADY in the queue = Jesteś JUŻ w kolejce.

res/values-pl/strings.xml
```diff
@@ -621,7 +621,7 @@
     <string name="Status">"Status:"</string>
     <string name="upload_size_failure">"Nie można przesłać tego obrazu, ponieważ jest zbyt duży. Spróbuj użyć innego obrazu."</string>
     <string name="CONTRIBUTE_PLASMA">"PRZEKAŻ PLAZMĘ"</string>
-    <string name="Specify_Contribution_Amount">"Określ ilość przesyłanej Plazmy:"</string>
+    <string name="Specify_Contribution_Amount">"Podaj ilość plazmy do przekazania:"</string>
```
AnkaKoxu, line 624: Never have I seen in polish strings such translation: Podaj for Specify. it is translated in all strings to Określ in every single string.

AnkaKoxu (commented):
That translation has good some improvements and does make sense, however it needs to be corrected:

res/values-pl/strings.xml
Rom-hub-77: Should I fix the listed errors and recreate this file?

AnkaKoxu: **Developer/moderator always decides to make changes.** 
_It would be better to correct this, for the reasons mentioned above. This may increase the chance of merging._

PR #1523, opened 2025-11-01
Update strings.xml

dentroyah:
ID:19952459


John-Ozz:
No need to rename file, you could just change the lines and save them with the same file name (`strings.xml`).

MakotoJohn (approved):
The file was (and remains) accidentally renamed from `strings.xml` to `strings.xml 1087,1090`, but the changes look good.

PR #1527, opened 2025-11-08
Fix Polish translations in strings.xml

Karfen22:
Updated Polish translations for various strings in the app.


Karfen22:
I made necessary changes, for example, the ways in which we speak the language

AnkaKoxu:
Invalid changes

PR #1533, opened 2025-12-01
Update strings.xml

Adriace76:
ID: 24010036
Razón: el código "Comienza En" la palabra "En" debe ser minúscula, mi propuesta es "Comienza en".


MakotoJohn:
Can close; change is not relevant, titles are used to have capitals in each word, plus the English value is the same thing.

PR #1545, opened 2025-12-04
Update Portuguese translations

tarsomonrroy:
String translations updated for greater clarity and accuracy.

Dot Sommelier: from "Escançador de Pontos" to "Supremacia de Pontos", "escançador" is an archaic word rarely used specifically for "aquele que serve vinho, um copeiro", derived from "escançar" (to serve wine, to decant, to aerate).

Doge Pound: from "Libra dos Doges" to "Canil dos Doges", "Pound" was translated literally as "Libra", but in the pun it translates as "Canil".

Signing out will disconnect you from the current game: from "Deslogando irá desconectá-lo do jogo atual!" to "Deslogar irá desconectá-lo do jogo atual!", the meaning of the phrase.

Blob Chat Bubbles: from "Bolhas do Chat no Blob" to "Balão de Chat no Blob", in UI/UX for Portuguese when the message appears visually in this way it is called "balão" instead of "bolha".

Winner Winner Chicken Dinner: from "Winner Winner Chicken Dinner!" to "Virou Realeza!", an adaptation of the term to Portuguese referring to winning a "Batalha Real".

### DIFF
--- a/res/values-pt-rBR/strings.xml
+++ b/res/values-pt-rBR/strings.xml
@@ -153,7 +153,8 @@
     <string name="Dot_Gobbler">"Guloso de Pontos"</string>
     <string name="Dot_Devourer">"Devorador de Pontos"</string>
     <string name="Dot_Gormandizer">"Gordice de Pontos"</string>
-    <string name="Dot_Sommelier">"Escançador de Pontos"</string>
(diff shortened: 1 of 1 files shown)


MakotoJohn (approved):
The changes are good but you shouldn't have changed the strings position to avoid confusion, like you did in the score-related ones.

PR #1553, opened 2025-12-28
Español ES mejorar traducción

NextGamer98:
mi ID 715896

### DIFF
--- a/res/values-es/strings.xml
+++ b/res/values-es/strings.xml
@@ -1414,7 +1414,7 @@ Comandos:
 
 Si no deseas registrarte, aún puedes jugar en modo individual."
     </string>
-    <string name="_2x_skin_resolution">Resolución de skin doble</string>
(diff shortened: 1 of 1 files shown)


res/values-es/strings.xml
```diff
@@ -1414,7 +1414,7 @@ Comandos:
 
 Si no deseas registrarte, aún puedes jugar en modo individual."
     </string>
-    <string name="_2x_skin_resolution">Resolución de skin doble</string>
+    <string name="_2x_skin_resolution">Resolución de skin Doble</string>
```
MakotoJohn, line 1417: This change looks weird. As a non-native, here are my suggestions:

### Suggestion 1

```suggestion
    <string name="_2x_skin_resolution">Doble Resolución de Skin</string>

MakotoJohn: ### Suggestion 2
```suggestion
    <string name="_2x_skin_resolution">Doble Calidad de Skin</string>

PR #1554, opened 2025-12-29
Update strings.xml Spanish

berlinmultijugador-hash:
ID:** 33377925
Plasma ganado It is incorrect in spanish, it should be "Plasma ganada" because it is a feminine word in the Spanish language.


berlinmultijugador-hash (commented):
<string name="Plasma_Won">Plasma ganada</string>
Forma correcta.

MakotoJohn (commented):
Can close, there are too many inconsistent changes, especially the change between "Sí" (yes) and "Si" (if), which is grammatically inaccurate.

PR #1562, opened 2026-01-02
Update strings.xml

27544270:
Değişiklikler küçük ama yine de biraz yardımcı olmak istedim. ID 27544270


MakotoJohn:
I don't know if you noticed, but you attempted to change the Azerbaijani language, not Turkish. If you are unsure about the corresponding strings for Turkish, please check out [here](https://github.com/simplicialsoftware/nebulous-translations/blob/main/res/values-tr/strings.xml).

Although both languages are similar, the acronyms should help you identify which language is what you are looking for; `az` belongs to Azerbaijani and `tr` belongs to Turkish.

MakotoJohn:
Can close.

PR #1583, opened 2026-01-25
Update Malay translations in strings.xml

andyfeez7777:
ID: 23164256

### DIFF
--- a/res/values-ms/strings.xml
+++ b/res/values-ms/strings.xml
@@ -11,7 +11,7 @@
     <string name="DEFAULTS">"Lalai"</string>
     <string name="WEEKLY">"Mingguan"</string>
     <string name="DAILY">"Keharian"</string>
-    <string name="Server_">"Server:"</string>
+    <string name="Server_">"Servers:"</string>
     <string name="Game_Mode_">"Mod permainan:"</string>
     <string name="Show_Names">"Tunjuk Nama"</string>
     <string name="Show_Avatars">"Tunjuk Kulit"</string>
@@ -27,7 +27,7 @@
     <string name="Are_You_Sure_">"Adakah kamu pasti?"</string>
     <string name="Yes">"Ya"</string>
     <string name="No">"Tidak"</string>
-    <string name="WAIT">"TUNGGU…"</string>
+    <string name="WAIT">"TUNGGU"</string>
     <string name="Must_Specify_Name_">"Perlu meletak nama."</string>
 
     <string name="Content_rating_of_the_app_may_change_when_finding_groups_">"Kedudukan kandungan aplikasi mungkin berubah apabila mencari kumpulan."</string>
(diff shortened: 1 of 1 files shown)


res/values-ms/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="DEFAULTS">"Lalai"</string>
     <string name="WEEKLY">"Mingguan"</string>
     <string name="DAILY">"Keharian"</string>
-    <string name="Server_">"Server:"</string>
+    <string name="Server_">"Servers:"</string>
```
MakotoJohn, line 14: As far as I know, Indonesian and Malay do not distinguish between plurals, plus there's not needed to pluralise the word at all, because it's not about **server count or list**, but about **server selection**.

res/values-ms/strings.xml
```diff
@@ -27,21 +27,21 @@
...
 
     <string name="_Rank_">" Pangkat "</string>
     <string name="_Alias_">" Alias "</string>
     <string name="_Score_">" Skor "</string>
-    <string name="Score__">"Skor: "</string>
-    <string name="Recombine_">"Bergabung: "</string>
-    <string name="WINNER_">"PEMENANG:"</string>
+    <string name="_Score_">"Skor: "</string>
+    <string name="_Recombine_">"Bergabung: "</string>
+    <string name="_WINNER_">"PEMENANG:"</string>
```
MakotoJohn, line 40: This should **never** be changed, otherwise the strings would not work as expected; they would be untranslated _if developer accidentally merges these changes into_ `main`.

res/values-ms/strings.xml
```diff
@@ -27,21 +27,21 @@
...
     <string name="IDLE">"TERBIAR"</string>
     <string name="Name_Invalid_">"Nama tidak sah."</string>
-    <string name="Name_Taken_">"Nama diambil."</string>
-    <string name="CONNECTED">"DIKAIT"</string>
+    <string name="Name_Taken_">"Nama di ambil."</string>
+    <string name="CONNECTED">"DI KAIT"</string>
```
MakotoJohn, line 44: I've researched about these translations and both are incorrect.

```suggestion
    <string name="Name_Taken_">"Nama Diambil."</string>
    <string name="CONNECTED">"TERHUBUNG"</string>

File: res/values-ms/strings.xml [hunk 4]
```diff
@@ -50,7 +50,7 @@
 
 
     <string name="Lobby_Name">"Nama lobi"</string>
-    <string name="Not_found_">"Tidak ditemui."</string>
+    <string name="Not_found_">"Tidak di temui."</string>
```
MakotoJohn, line 53: This one should also not be changed because the previous translation was correct.

res/values-ms/strings.xml
```diff
@@ -95,7 +95,7 @@
     <string name="SIGN_OUT">"TANDA KELUAR"</string>
     <string name="ACHIEVEMENTS">"CAPAIAN"</string>
     <string name="STATISTICS">"STATISTIK"</string>
-    <string name="XP_">"XP:"</string>
+    <string name="XP_">"XP;"</string>
```
MakotoJohn, line 98: The change to a semicolon does _not_ make sense in this case.

MakotoJohn:
Can close and skip to #1588.

PR #1584, opened 2026-01-25
Portugues BR Translation Improvements

victo7pxy:
7357702


MakotoJohn:
As a native, I only agree with the changes at lines 1223 and 1226, the other rest was actually unnecessary. Here is why:

- **Capitalisation:** The capitalisation is essential for titles in UI, and it's advised for it to be followed based on original (English).
- **Change from "vs" to "x":** The "vs" already clarifies that it's _versus_ someone, changing it to _x_ looks like the system is talking about size, actually.

PR #1591, opened 2026-01-29
Update Arabic translations in strings.xml

lohdi:
ID: 31463743


TurkiAlanazi1:
The changes are incorrect and this request should not be accepted.

PR #1598, opened 2026-02-01
Finnish FI Translation Improvements

veetikemppainen:


res/values-fi/strings.xml
```diff
@@ -1352,7 +1352,7 @@ Commands:
         <item>"Pieni"</item>
         <item>"Normaali"</item>
         <item>"Suuri"</item>
-        <item>"HUGE"</item>
+        <item>"Valtava"</item>
```
MakotoJohn, line 1355: It must remain 100% capital.

```suggestion
        <item>"VALTAVA"</item>

PR #1606, opened 2026-02-05
Refactor Portuguese strings for clarity and updates

lorddeath098abc-dotcom:
Updated various string resources for better clarity and consistency in the Portuguese (Brazil) localization.

Id: 26692366 

NEBULOUS


MakotoJohn:
No need of `</resources>` there; other rest will remain untranslated. Also, if you are actually going to add _comments_ to organise the strings, you might have to work harder than this.

PR #1613, opened 2026-02-06
Fix typo in winner string in Italian localization

sh7276182-a11y:
ID 32258047


AnkaKoxu:
Duplicated from #1225

PR #1627, opened 2026-02-27
Update strings.xml 🇹🇷

KagaN4469:
ID : 22668744


KagaN4469 (commented):
Translations ID : 22668744

PR #1652, opened 2026-04-15
Update strings.xml

jadalmrashdeh-hue:
ID:15580467


jadalmrashdeh-hue:
thanks

TurkiAlanazi1:
Incorrect changes and the request should not be accepted.

PR #1686, opened 2026-06-06
Vietnamese Translation Improvements

trankienvaae:
ID: 24892649


trankienvaae:
> ID: 24892649
Correct the stiff wording

PR #1713, opened 2026-07-19
Update strings.xml

mohammedkazim8008213714-sudo:
Id : 22289139


mohammedkazim8008213714-sudo (commented):
Changed few language hindi from India, looking for translator tag Id : 22289139

PR #1726, opened 2026-08-15
Fix translations in strings.xml

XANZ48:
9. Buttons > Tombol
17. Play > Main

My ID 10055954


XANZ48:
17. Play > Main (not on number 10)

